PR #3, opened 2021-05-19
Add graphql, queries, and mutations

gaoxk:
## Notion ticket link

[Add GraphQL and create example mutation](https://www.notion.so/uwblueprintexecs/Add-GraphQL-and-create-example-mutation-4c2eb3b541504533995cddc7302a44da)
[GraphQL Migration Epic](https://www.notion.so/uwblueprintexecs/GraphQL-Migration-261d7b97ce8e455abeb1859b217898af)

## Implementation description
* Add graphql service ontop of flask
* Add allUsers and allEntities queries
* add createEntities mutation 
* graphql interactive ui available at localhost:5000/graphql

Note, the REST api still exists and functions. The frontend should still operate as expected. 

## Steps to test
1. build and run
2. go to localhost:5000/graphql
3. test out: 
```
mutation {
  createUser(userData:{firstName:"vov", lastName:"rob", email:"asddf@asdf.com", password: "asdfffdf", role: User}){
    user{
      id
    }
  }
}
```
```
mutation {
  createEntity(entityData:{stringField:"AS", intField:1, stringArrayField:["a"], boolField:true, enumField:A}){
    entity{
      id
    }
  }
}
```
```
{
	users {
		id
  }
}
```
```
{
	entities {
		id
  }
}
```

**Not sure if my pattern for resolvers + types makes sense. Please roast brutally.** 

## Checklist
- [ - ] My PR name is descriptive and in imperative tense
- [ - ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ - ] I have run the appropriate linter(s)
- [ - ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/python/app/services/implementations/entity_service.py
```diff
@@ -1,56 +1,57 @@
...
-
\ No newline at end of file
+from ...models.entity import Entity
+from ...models import db
+from ..interfaces.entity_service import IEntityService
+
+from flask import current_app
+
+class EntityService(IEntityService):
+    def __init__(self, logger=current_app.logger):
```
gaoxk, line 8: git diff looks weird here. This line + flask import are the only changes.

backend/python/app/graphql/types/entity_type.py
```diff
@@ -0,0 +1,35 @@
...
+        interfaces = (relay.Node, )
+
+class EntityInput(graphene.InputObjectType):
+    string_field = graphene.String(required=True)
+    int_field = graphene.Int(required=True)
+    string_array_field = graphene.List(graphene.String, required=True)
+    enum_field = graphene.String(required=True) # TODO: make stricter
+    bool_field = graphene.Boolean(required=True)
+
+def entity_input_to_dto_and_model(entity_data):
```
gaoxk, line 20: Please roast this pattern, I have no clue what I'm doing. 
I convert entity_data to entity_model as the mutation throws a fit if the returned object lacks an `id` field. The typescript graphql configuration of the starter-code has `EntityRequestDTO` and `EntityResponseDTO`, which is a potential solution, though `Entity` (model) already exists. LMK what you think.

backend/python/app/models/__init__.py
```diff
@@ -3,6 +3,9 @@
 db = SQLAlchemy()
 erase_db_and_sync = False
 
+from .user import User
+from .entity import Entity
+
```
lynn-kim, line 8: Can either one of the sets for the User/Entity imports be removed?

backend/python/app/graphql/types/entity_type.py
ZhangCreations, line 20: A summary of Kristy's comment which I accidentally deleted: Not sure if this is how it should be implemented. Currently doing it this way since typescript wouldn't work without an id. Starter code has `EntityDTORequest` and `EntityDTOResponse` to solve this issue, but it might be redundant for us since we already have `Entity (model)`

ZhangCreations: In the past I've just interacted directly with the model instead of using a DTO with GraphQL and based on what I read in a quick google search of "GraphQL DTO" it sounds like the two design patterns go against each other.

backend/python/app/models/__init__.py
gaoxk: It kiiinda helps with imports. With these two lines, this is allowed:
```
from ...models import Entity as EntityModel
```
Without these lines you have to do:
```
from ...models.entity import Entity as EntityModel
```
honestly, now that I look at it more, I think the latter is more clear. I'm removing these lines now.

backend/python/app/graphql/types/entity_type.py
gaoxk: 👍 read up on DTO, will attempt to ditch the DTO all together and update the pr.

backend/python/app/services/implementations/entity_service.py
sherryhli: Just to provide the background on this, the intention was to limit the flask dependency to only the controller layer - no issues with this though! 🙂

backend/python/app/graphql/resolvers/entity_resolver.py
```diff
@@ -0,0 +1,21 @@
...
+
+class CreateEntity(graphene.Mutation):
+    class Arguments:
+        entity_data = EntityInput(required=True)
+        
+    ok = graphene.Boolean()
+    entity = graphene.Field(lambda: Entity)
+
+    def mutate(root, info, entity_data=None):
+        entity_dto, entity_model = entity_input_to_dto_and_model(entity_data)
```
sherryhli, line 14: Are both the DTO and model necessary here? I think ideally the model should be hidden from the resolver and the resolver should deal with DTOs only, otherwise this defeats the purpose of having the service layer.

Disclaimer that I have not worked with GraphQL extensively but I think it would be beneficial to decouple the model (and by extension the ORM) from the GraphQL resolver.

Would line 17 work if you pass in the result of line 15?
```python
# I think it returns a dict right now but it essentially functions as an "EntityResponseDTO"
services["entity"].create_entity(entity_dto)
```

backend/python/app/graphql/types/entity_type.py
sherryhli: ^ Yep seems like there's a bit of mix and match happening, see my comment in `backend/python/app/graphql/resolvers/entity_resolver.py`

Starter code hides the ORM from the controller layer (i.e. REST endpoints or GraphQL resolvers) by using service classes which deal with DTOs instead of the underlying model. If you'd like to keep this pattern then I would recommend not using `graphene_sqlalchemy`. If using `graphene_sqlalchemy` then I don't think the service layer (plus DTOs) will be necessary at all.

I don't know what Python + GraphQL best practices are, personally biased towards first approach for obvious reasons 😛 but totally up to your team to decide!

sherryhli (commented):
Added some thoughts but definitely feel free to disregard, 100% your team's decision 🙂

Also should have documented this somewhere but you can use the `black` formatter:
```bash
docker exec -it <backend-container-name> /bin/bash -c "black ."
```

backend/python/app/graphql/types/entity_type.py
gaoxk: ah I ended up flip flopping on this lol: DTOs are much easier to work with for things like Users where emails and passwords arent part of the ORM, but auth_id is hidden. DTOs are back in for now and `graphene_sqlalchemy` has been removed. 

I'm leaving in `/resources/models` for now as we can remove them later when we do the full REST nuke. 

Ty for the feedback!

gaoxk:
`graphene_sqlalchemy` has been ditched in favour of DTOs. CreateUser mutation is now also in the pr. Mutations and queries have been split to two different directories.

backend/python/app/graphql/types/entity_type.py
```diff
@@ -0,0 +1,16 @@
+import graphene
+
+class EntityRequestDTO(graphene.InputObjectType):
```
lynn-kim, line 3: Is the difference in naming for the Entity DTOs / User DTOs intentional?

lynn-kim (commented):
Left a few comments but otherwise LGTM!

backend/python/app/graphql/types/entity_type.py
gaoxk: I'm mainly following the patterns in `resources/`. As far as I can tell, UserDTO is the "response dto" for all user endpoints. Each user endpoint has its own "request dto" (i,e CreateUserDTO)

backend/python/app/graphql/types/entity_type.py
```diff
@@ -0,0 +1,16 @@
...
+    string_field = graphene.String(required=True)
+    int_field = graphene.Int(required=True)
+    string_array_field = graphene.List(graphene.String, required=True)
+    enum_field = graphene.String(required=True) # TODO: make stricter
```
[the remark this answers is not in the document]
gaoxk: Ah on further review I realize that Enums can be strings. Don't know why I thought they could only be stored as ints on the backend. Change has been made.

gaoxk:
TFTR! 🚢

PR #13, opened 2021-05-29
Migrate reset password endpoint to GraphQL

e-wai:
## Notion ticket link
[Migrate reset_password endpoint to GraphQL](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=b69224f1963f4ac195c62c71bff4f6da)

## Implementation description
* Add mutation for `resetUser(email: String!): ResetUser`
* Migrate call in frontend from REST to GraphQL
* Change `index.tsx` to meet Apollo requirements

## Steps to test
1. From backend, try registered email, non-registered emails in GraphQL playground
2. From frontend, try registered email, non-registered email, non-email strings in entry field and check expected message occurs

* Was it necessary to add `services["user"]` in the `__init__.py`?
* Why have other migrations not run into the apollo client issue that required changes in `index.tsx`? ✅ (just hadn't had any other front-end changes yet that were trying to import)
* Am I using Graphene incorrectly -> can we do `resetUser(email: String!): Boolean!` instead of `resetUser(email: String!): ResetUser`?

## Checklist
- [x] I have run the appropriate linter(s)

### DIFF
--- a/backend/python/app/graphql/mutations/auth_mutation.py
+++ b/backend/python/app/graphql/mutations/auth_mutation.py
@@ -4,4 +4,41 @@ TODO mutations:
     refresh: String!
     logout(userId: ID!): ID
     resetPassword(email: String!): Boolean!
- """
+"""
+
+import os
+
+import graphene
+from flask import Blueprint, current_app, jsonify, request
+
+from ...services.implementations.auth_service import AuthService
+from ...services.implementations.email_service import EmailService
+from ..service import services
+
+email_service = EmailService(
+    current_app.logger,
+    {
+        "refresh_token": os.getenv("EMAIL_REFRESH_TOKEN"),
+        "token_uri": "https://oauth2.googleapis.com/token",
+        "client_id": os.getenv("EMAIL_CLIENT_ID"),
+        "client_secret": os.getenv("EMAIL_CLIENT_SECRET"),
+    },
+    "planetread@uwblueprint.org",  # must replace
+    "Planet Read",  # must replace)
+)
+auth_service = AuthService(current_app.logger, services["user"], email_service)
+
+
+class ResetPassword(graphene.Mutation):
+    class Arguments:
+        email = graphene.String(required=True)
+
+    ok = graphene.Boolean()
+
+    def mutate(root, info, email):
+        try:
+            auth_service.reset_password(email)
+            return ResetPassword(ok=True)
+        except Exception as e:
+            error_message = getattr(e, "message", None)
+            raise Exception(error_message if error_message else str(e))
--- a/frontend/src/components/auth/ResetPassword.tsx
+++ b/frontend/src/components/auth/ResetPassword.tsx
@@ -1,10 +1,18 @@
 import React, { useContext } from "react";
-import authAPIClient from "../../APIClients/AuthAPIClient";
+import { gql, useMutation } from "@apollo/client";
 import AuthContext from "../../contexts/AuthContext";
 
 interface ResetPasswordProps {
   email?: string;
 }
+
+const RESET_PASSWORD = gql`
+  mutation ResetPassword($email: String!) {
+    resetPassword(email: $email) {
+      ok
+    }
+  }
+`;
 
 const ResetPassword = ({ email }: ResetPasswordProps) => {
   const { authenticatedUser } = useContext(AuthContext);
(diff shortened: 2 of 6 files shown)


frontend/src/components/auth/ResetPassword.tsx
```diff
@@ -14,17 +22,23 @@ const ResetPassword = ({ email }: ResetPasswordProps) => {
     return re.test(String(emailString).toLowerCase());
   };
 
+  type ResetPassword = { ok: boolean };
+  const [resetPassword] = useMutation<{ resetPassword: ResetPassword }>(
```
gaoxk, line 26: > Am I using Graphene incorrectly -> can we do resetUser(email: String!): Boolean! instead of resetUser(email: String!): ResetUser?

For context: Emma tried `useMutation<{ ok: boolean }>` but ran into this problem:
```
let res = await resetPassword({ variables: { email: resetEmail } });
console.log(res);
// {data: resetPassword: {ok: true, __typename: "ResetPassword"}}
console.log(res.data.resetPassword);
// Property 'resetPassword' does not exist on type '{ ok: boolean; }'
```
The conclusion is that `useMutation` sets what Typescript expects res to look like. While Emma's code works, the pattern feels odd. `Logout.tsx` in the graphql starter-code doesnt have to follow something like this.

@lynn-kim @kevinwang2040 What have y'all been experiencing? I don't have a good answer for why this is going on.

frontend/src/components/auth/ResetPassword.tsx
```diff
@@ -14,17 +22,23 @@ const ResetPassword = ({ email }: ResetPasswordProps) => {
...
     if (authenticatedUser == null && !isRealEmailAvailable(email!)) {
-      alert("invalid email");
+      alert("Invalid email");
       return;
     }
+
     const resetEmail = authenticatedUser?.email ?? email;
-    if (await authAPIClient.resetPassword(resetEmail)) {
+    try {
+      await resetPassword({ variables: { email: resetEmail } });
```
gaoxk, line 38: `ok: false` could come back here, and we'd still be alerting users that a reset email was sent. Can you add a check here? This should work right now:
```
const res = await resetPassword({ variables: { email: resetEmail } });
console.log(res.data?.resetPassword.ok);
```
I know we don't have an `ok: false` case right now, but I'd rather cover our bases.

backend/python/app/graphql/mutations/auth_mutation.py
```diff
@@ -4,4 +4,41 @@
...
+
+
+class ResetPassword(graphene.Mutation):
+    class Arguments:
+        email = graphene.String(required=True)
+
+    ok = graphene.Boolean()
+
+    def mutate(root, info, email):
+        try:
```
gaoxk, line 39: I think this try statement is unnecessary. An error raised in `auth_service.reset_password(email) ` will continue to be raised down the call stack, so I don't think you need to re-raise this `except`

gaoxk (changes requested):
looking good! feel free to start have the conversation about useMutation with Kevin and Lynn on slack. 

> Was it necessary to add services["user"] in the __init__.py?

yep! The services need the app logger, which isnt available until the flask app has started.graphql/__init__.py runs after the flask app has started.

frontend/src/components/auth/ResetPassword.tsx
e-wai: The pattern in the frontend is a result of how the GraphQL mutation responds, ie. `resetUser(email: String!): ResetUser` instead of `resetUser(email: String!): Boolean!`? -- so I think it's a limitation of how Graphene has mutations and I haven't found a workaround

Related Q: https://github.com/graphql-python/graphene/issues/497

backend/python/app/graphql/mutations/auth_mutation.py
e-wai: Hmm this is what the REST endpoint looks like, what should this be? Only keeping lines 40-41 without a `try` wrapping?

gaoxk: huh. You're right, that is how the REST endpoint does it. Feel free to keep the same code, I'm going to ask internal tools about this choice.

frontend/src/components/auth/ResetPassword.tsx
gaoxk: As an experiment, you could try copying your `ResetPassword.tsx` (and other necessary files/changes) to the graphql configured starter-code and seeing what works. Im curious if the typescript graphql server from starter-code also follows the `data:resetuser:ok` pattern.

e-wai: Their frontend looks like:
```
const RESET_PASSWORD = gql`
  mutation ResetPassword($email: String!) {
    resetPassword(email: $email)
  }
`;

const [resetPassword] = useMutation<{ resetPassword: boolean }>(
  RESET_PASSWORD,
);

const onResetPasswordClick = async () => {
  await resetPassword({ variables: { email: authenticatedUser?.email } });
};
```

which matches their backend, where the mutation is `resetPassword(email: String!): Boolean!` and they don't have to filter through a `ResetPassword` object to get the result?

gaoxk: rip

gaoxk (approved):
🚢 🚢 🚢 🚢

PR #11, opened 2021-05-28
Migrate get_users endpoint to GraphQL resolver

jennchenn:
## Notion ticket link
[Migrate get_users endpoint to GraphQL resolver](https://www.notion.so/uwblueprintexecs/Migrate-get_users-endpoint-to-GraphQL-resolver-460b7d7bb2b0429190163f6f7baf954d)

## Implementation description
* add `resolve_user_by_id` and `resolve_user_by_email` to `graphql/queries/user_query.py`
* add calls to `resolve_user_by_id` and `resolve_user_by_email` to `graphql/schema.py`

## Steps to test
1. Start up the backend
2. Ensure the database has existing users; otherwise seed database with users:
```
$ planet-read_db_1 psql -U postgres -d planet-read
$ \dt
```
>> if no tables present, change the `erase_db_and_sync` = `False` to `True` in `backend/python/app/models/__init__.py`
```
$ SELECT * FROM users;
```
>> if empty, insert users into the database
```
$ INSERT INTO users (first_name, last_name, auth_id, role) VALUES ('First', 'Last', 'insert-firebase-uid', 'Admin');
```
3. Go to `localhost:5000/graphql` and verify the response of the following requests returns the corresponding users that were queried:
```
{
  userById(id: <valid id of user as int>) {
    id
    firstName
    lastName
    role
    email
  }
  userByEmail(email: <valid email address of a user as string>) {
    id
    firstName
    lastName
    role
    email
  }
}
```

## What should reviewers focus on?
* `graphql/queries/user_query.py`
* `graphql/schema.py`
* files were formatted with black--is this the linter we want to use? it added a bunch of spaces+line breaks throughout the file

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #8, opened 2021-05-23
Store and retrieve files

jennchenn:
## Notion ticket link
[Store and retrieve files](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=9352edb12bb8460294b9b505d36ee523)

## Implementation description
* Added Files model (id, path)
* Added POST /files endpoint to allow users to upload a single file
* Added GET /files/<id> to allow users to retrieve path to file
* Files are uploaded to a folder in the container and not to the cloud at the moment

## Steps to test
1. Add an `UPLOAD_PATH=/var/lib/file_uploads` environment variable to point to the directory that files will be uploaded to 
2. Tear down containers and rebuild:
```
$ docker-compose down --volumes
$ docker-compose up --build
```
3. Go into `/backend/python/app/models/__init__.py` and change the `erase_db_and_sync = False` to `True`
4. Verify the Files table is created:
```
$ docker exec -it <db-container-id> psql -U postgres -d planet-read
$ \dt
```
image

5. Login and obtain the access_token for a user
```
curl --location --request POST 'localhost:5000/auth/login' \
--header 'Content-Type: application/json' \
--data-raw '{
    "email": <USER EMAIL ADDRESS>,
    "password": "<USER PASSWORD>
}'
```

6. Try to upload a file to the database (use the access_token from above):
```
curl --location --request POST 'http://localhost:5000/files' \
--header 'Authorization: Bearer  <ACCESS_TOKEN>' \
--form 'file=@"<PATH_TO_FILE>"'
```
Observe that the response looks something like:
```
{
    "id": 1,
    "path": "./app/files/my_file.pdf"
}
```

7. Try to retrieve the file path from the database:
```
curl --location --request GET 'http://localhost:5000/files/1' \
--header 'Limit: 1' \
--header 'Authorization: Bearer <ACCESS_TOKEN>' \
--header 'Content-Type: application/json' 
```
Observe that the response looks something like:
```
{
    "id": 1,
    "path": "./app/files/my_file.pdf"
}
```
8. Verify that the files are stored correctly in the volume:
```
$ docker run -it --rm -v planet-read_file_uploads:/vol busybox ls -l /vol
```

## What should reviewers focus on?
* Verifying that the endpoints are structured correctly
* Actual file changes (ran the linter so the git diff looks really messy sorry about that!):
  * `backend/python/app/models/file.py`
[...]

PR #14, opened 2021-06-03
Add authorization for user GraphQL endpoints

e-wai:
## Notion ticket link
[Add authorization checks to Graphql resolvers](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=da8927857a284356b83d10053b9ff230)

## Implementation description
* Added new middleware method `require_authorization_by_role_gql` that performs the same checks as its non-GraphQL equivalent. Needed to change the arguments and how it accessed headers (request object is in `info.context` rather than just `request`)
* Added the decorator to the appropriate mutations and queries. Justification for adding it on the business logic layer rather than the GraphQL layer (in `schema.py`) based on https://graphql.org/learn/authorization/ 

Also, I added the method into the existing `auth.py` instead of creating a new file because when we eliminate the REST code, the file name still makes sense and now requires a single method name refactor.

## Steps to test

Setup -> I added users manually (see https://github.com/uwblueprint/planet-read/pull/11). Not really necessary, but makes for slightly more interesting testing 🤷 

1. NOT AUTHORIZED case: Open `localhost:5000/graphql` and mess around with any of the queries and any of the mutations except for `resetPassword`. Should respond that "You are not authorized to make this request.", eg. below. Can also produce this in Postman if you have no/an incorrect `Authorization` header.
```
{
  "errors": [
    {
      "message": "You are not authorized to make this request.",
      "locations": [
        {
          "line": 2,
          "column": 2
        }
      ],
      "path": [
        "entities"
      ]
    }
  ],
  "data": {
    "entities": null
  }
}
```
2. AUTHORIZED case: Find access token with `console.log(window.localStorage.getItem("localhost:AUTHENTICATED_USER"));` in front-end console. Use this (formatted as `Bearer <access-token>`) as the `Authorization` header and try making any of the above failing requests. Should return as expected, no errors. I used Postman for this.

* Should this ticket include a GraphQL verison of `require_authorization_by_user_id`? No way to test without logout GraphQL endpoint being established
* I don't know how to accomplish the headers in GraphiQL so I used Postman instead. I know Postman can hide CORS errors -- is this a flawed way of testing?

## Checklist

### DIFF
--- a/backend/python/app/graphql/mutations/entity_mutation.py
+++ b/backend/python/app/graphql/mutations/entity_mutation.py
@@ -1,5 +1,6 @@
 import graphene
 
+from ...middlewares.auth import require_authorization_by_role_gql
 from ..service import services
 from ..types.entity_type import EntityRequestDTO, EntityResponseDTO
 
@@ -11,6 +12,7 @@ class CreateEntity(graphene.Mutation):
     ok = graphene.Boolean()
     entity = graphene.Field(lambda: EntityResponseDTO)
 
+    @require_authorization_by_role_gql({"User", "Admin"})
     def mutate(root, info, entity_data=None):
         entity_response = services["entity"].create_entity(entity_data)
         ok = True
--- a/backend/python/app/graphql/mutations/user_mutation.py
+++ b/backend/python/app/graphql/mutations/user_mutation.py
@@ -1,5 +1,6 @@
 import graphene
 
+from ...middlewares.auth import require_authorization_by_role_gql
 from ..service import services
 from ..types.user_type import CreateUserDTO, UserDTO
 
@@ -11,6 +12,7 @@ class CreateUser(graphene.Mutation):
     ok = graphene.Boolean()
     user = graphene.Field(lambda: UserDTO)
 
+    @require_authorization_by_role_gql({"User", "Admin"})
     def mutate(root, info, user_data=None):
         user_response = services["user"].create_user(user_data).__dict__
         ok = True
--- a/backend/python/app/graphql/queries/entity_query.py
+++ b/backend/python/app/graphql/queries/entity_query.py
@@ -1,5 +1,7 @@
+from ...middlewares.auth import require_authorization_by_role_gql
 from ..service import services
 
 
+@require_authorization_by_role_gql({"User", "Admin"})
 def resolve_entities(root, info, **kwargs):
     return services["entity"].get_entities()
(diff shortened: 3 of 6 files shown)


e-wai:
Did some front-end testing by adding `@require_authorization_by_role_gql({"Admin"})` to wrap this method

https://github.com/uwblueprint/planet-read/blob/01e8a7f9aa64c590cd40a0662a71dc033c96af16/backend/python/app/graphql/mutations/auth_mutation.py#L38-L44

+ I expected that this would produce `Error: You are not authorized to make this request.` when not logged in or when logged in as an user, and complete successfully if logged in as an admin.
+ Behaviour when logged out was fine. When logged in as admin and as user, request always returned bad request (`{"errors":[{"message":"Cannot query field \"refresh\" on type \"Mutation\".","locations":[{"line":3,"column":5}]}]}`)

gaoxk (approved):
Nice! 

Fun fact: you can also get the jwt token via curl request, see Jenn's file PR (#8 )

Regarding using Headers for graphql requests, unfortunately, you're correct that graphiql doesn't support this. As far as I know, Postman should be fine. Personally, I installed Altair on to chrome and set the header `Authorization` to `Bearer <token>` with no problems. 

Let's skip `require_authorization_by_user_id` for now.

gaoxk:
Can you docker-compose down and up and try the front-end testing with `resetPassword` again? I tested the graphql endpoint and couldnt recreate the issues
`reset_password` doesn't require authorization so we don't need to worry about this too much, but it's good to double check. I also suspect this has something to do with the `refresh` gql mutation being incomplete, so we may resolve the problem relatively soon.

e-wai:
hm I tried again and got the same behaviour. The GraphQL endpoint wasn't causing the problem, it's that it's querying refresh so I can't fully test it. 

What happened for you when you clicked the "Reset Password" button when not logged in?

gaoxk:
Only tested the endpoint from the backend.  Good to know that this is an issue - I think you're good to rebase and squash and we can test again once the refresh endpoint is done. 
`{"errors":[{"message":"Cannot query field \"refresh\" on type \"Mutation\".","locations":[{"line":3,"column":5}]}]} this error occurs when graphql is unable to find the mutation named `refresh`. If you go into graphiql and give it a random mutation name, you'll get the same problem.

PR #22, opened 2021-06-09
Graphql migration/migrate refresh endpoint

gaoxk:
## Notion ticket link

[Migrate refresh endpoint to GraphQL ](https://www.notion.so/uwblueprintexecs/Migrate-refresh-endpoint-to-GraphQL-61748891a6b94ad294acec30a6aaf001)

## Implementation description
This one is a doozy
* use auth_service to refresh token in new refresh mutation 
* graphene mutations are wacko and cant send response header. The hack around this is
  * make a custom graphql view that wraps the [dispatch_request](https://github.com/graphql-python/graphql-server/blob/master/graphql_server/flask/graphqlview.py#L76) function 
  * read the refresh_token from the refresh mutation response and set it as a response cookie
* **the user MUST request for `refresh_token` in the refresh mutation**  

There were a few implementation options in the realm of custom gql view:
1. Completely hijack the refresh mutation and recompute `auth_service.renew_token`. two problems
  i. completely rewriting `response.response.data` breaks request encryption and makes the web app explode. The response from refresh looked correct, but the web app exploded with generic apollo errors. 
 ii. calling the auth service from the view was a very weird pattern
The body of dispatch_request looked more like this when trying to implement this idea:
```
token = auth_service.renew_token(request.cookies.get("refreshToken"))
access_token = token.access_token
refresh_token = token.refresh_token
data = {
      "refresh": {
         "accessToken": access_token,
                    "ok" : True
       }
}
response.set_cookie("refreshToken", refresh_token)
response.response = [json.dumps({'data':data}).encode('utf-8')]
```
2. `renew_token` within the refresh mutation and somehow pass up the `refresh_token` to the gql view
 i. impossible. The only way was to set class variables on the custom view itself, which will lead to terrible concurrency issues. `refresh_token` computed in the mutation was as good as gone

Regarding setting the response token on the server-side, I dislike the idea of leaving it to the client and the client cookie becoming stale because they forgot to set the cookie. 
On the other hand, requiring the client to request for the refresh_token for the response cookie to be set is dumb. I'm open to opinions. 

## Steps to test
Do this through the frontend.
1. make user/login
2. inspect element and open network tab
3. hit refresh button
4. b e h o l d the response cookie
image
5. profit

* please be critical of the cookie setting pattern and tell me your thoughts!

## Checklist


ZhangCreations (approved):
Yeah, while it's not ideal it seems to be the standard way to implement it based on other ppl's implementations

backend/python/app/graphql/__init__.py
```diff
@@ -1,24 +1,42 @@
...
+            graphiql=True,
+            context={"session": db.session},
         ),
     )
 
     services["entity"] = EntityService(current_app.logger)
     services["story"] = StoryService(current_app.logger)
     services["user"] = UserService(current_app.logger)
+
+    email_service = EmailService(
```
gaoxk, line 31: service changes are purely for code organization

backend/python/app/graphql/mutations/auth_mutation.py
```diff
@@ -52,20 +45,29 @@ class Arguments:
         email = graphene.String(required=True)
         password = graphene.String(required=True)
 
-    access_token = graphene.String(required=True)
+    access_token = graphene.String()
+    refresh_token = graphene.String()  # refresh_token MUST be requested
     id = graphene.Int()
-    first_name = graphene.String(required=True)
```
gaoxk, line 57: turns out these `required `fields don't do anything

gaoxk:
Logout and signup didnt set the response cookie previously - those are now also set. You can test by login in as a user and hitting the refresh button, and also by signing up as a user and hitting the refresh button

lynn-kim (approved):
lgtm!

PR #24, opened 2021-06-12
Create and Assign User to Story Translation

ZhangCreations:
## Notion ticket link
[Assign a User as translator to a story](https://www.notion.so/uwblueprintexecs/Assign-a-User-as-translator-to-a-story-5dbab72c1a53461c96dcf39a7ea24a64)

## Implementation description
* Created a new DTO for request & response for creating a story translation
* Creates a new entry for the story translation
* Fetches all the lines associated with the story that's being translated and makes empty lines for those
* Returns the initial story translation entry

## Steps to test
1. Build the backend container & go to the [GQL playground](http://localhost:5000/graphql)
2. Ensure you have a story and a user.
3. Query for a random user_id and story_id `stories{user_id}` and `users{user_id}` should do it, fetch any random numbers from those lists
4. Run the following mutation
```
mutation($storyTranslationData: NewStoryTranslationRequestDTO!){
              createStoryTranslation(storyTranslationData: {
                "translatorId": ${INSERT USER_ID  HERE},
                "storyId": ${INSERT STORY_ID  HERE},
                "language": "English"
            }){
                story{
                  id
                  translatorId
                  storyId
                  language
                  stage
		}
	}
}
```
5. Profit
6. Now to test error states, try any of the following: Invalid story_id/user_id/language/stage

* I don't like the code pattern because of the repeated db.commit but according to [this helpful article I only found after debugging for like 30 minutes](https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/) you only get the primary key of the object after you commit. So if anyone has a workaround to this lmk 👍
* Also, this means I used like double try except loops so we don't ever have a case where a query fails when making the StoryTranslationContents entries and then just leaves the instantiated StoryTranslation entry existing so, lmk if there's a workaround to this as well.

## Checklist

### DIFF
--- a/backend/python/app/graphql/types/story_type.py
+++ b/backend/python/app/graphql/types/story_type.py
@@ -33,3 +33,18 @@ class StoryRequestDTO(graphene.InputObjectType):
     youtube_link = graphene.String(required=True)
     level = graphene.Int(required=True)
     translated_languages = graphene.List(LanguageEnum)
+
+
+class NewStoryTranslationResponseDTO(graphene.ObjectType):
+    id = graphene.Int(required=True)
+    translator_id = graphene.Int(required=True)
+    story_id = graphene.String(required=True)
+    language = graphene.String(required=True)
+    stage = graphene.Field(StageEnum)
+
+
+class NewStoryTranslationRequestDTO(graphene.InputObjectType):
+    translator_id = graphene.Int(required=True)
+    story_id = graphene.Int(required=True)
+    language = graphene.String(required=True)
+    stage = graphene.Field(StageEnum, default_value=StageEnum.START)
--- a/backend/python/app/graphql/types/story_type.py
+++ b/backend/python/app/graphql/types/story_type.py
@@ -35,7 +35,7 @@ class StoryRequestDTO(graphene.InputObjectType):
     translated_languages = graphene.List(LanguageEnum)
 
 
-class NewStoryTranslationResponseDTO(graphene.ObjectType):
+class CreateStoryTranslationResponseDTO(graphene.ObjectType):
     id = graphene.Int(required=True)
     translator_id = graphene.Int(required=True)
     story_id = graphene.String(required=True)
@@ -43,8 +43,8 @@ class NewStoryTranslationResponseDTO(graphene.ObjectType):
     stage = graphene.Field(StageEnum)
 
 
-class NewStoryTranslationRequestDTO(graphene.InputObjectType):
+class CreateStoryTranslationRequestDTO(graphene.InputObjectType):
     translator_id = graphene.Int(required=True)
     story_id = graphene.Int(required=True)
     language = graphene.String(required=True)
-    stage = graphene.Field(StageEnum, default_value=StageEnum.START)
+    stage = graphene.Field(StageEnum, default_value="TRANSLATE")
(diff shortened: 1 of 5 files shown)


backend/python/app/services/implementations/story_service.py
```diff
@@ -35,6 +37,7 @@ def create_story(self, story, content):
         db.session.commit()
 
         # insert contents into story_contents
+        # TODO use batch inserts here instead of iterating over the for loop?
```
ZhangCreations, line 40: Just a potential fix 🤷

backend/python/app/graphql/types/story_type.py
```diff
@@ -33,3 +33,18 @@ class StoryRequestDTO(graphene.InputObjectType):
     youtube_link = graphene.String(required=True)
     level = graphene.Int(required=True)
     translated_languages = graphene.List(LanguageEnum)
+
+
+class NewStoryTranslationResponseDTO(graphene.ObjectType):
```
gaoxk, line 38: nit: can we name these `CreateStoryTranslationResponseDTO` etc. Keeps in convention with other creation DTOS

backend/python/app/models/story_translation.py
```diff
@@ -10,10 +10,11 @@
 
 
 class StoryTranslation(db.Model):
+
     __tablename__ = "story_translations"
     id = db.Column(db.Integer, primary_key=True, nullable=False)
     story_id = db.Column(db.Integer, db.ForeignKey("stories.id"), nullable=False)
-    language = db.Column(pgEnum(LanguageEnum), nullable=False)
```
gaoxk, line 16: Why change to string column?

backend/python/app/services/implementations/story_service.py
gaoxk: good suggestion 👍

backend/python/app/services/implementations/story_service.py
```diff
@@ -51,3 +54,33 @@ def create_story(self, story, content):
         db.session.refresh(new_story)
 
         return new_story
+
+    def create_translation(self, entity):
+        try:
```
gaoxk, line 59: I'm on the fence about the nested try-excepts. What do you think about one try statement for the translation, and one try statement for the conents? if the contents throws, you can still delete the translation in the second except. 

Open to the nested try statements tho.

gaoxk (changes requested):
Looks good so far! Regarding double commits - there's no work around as you need the storytranslation id to make the storytranslationcontent.

gaoxk:
one more thing - can the creation mutation always set the stage to `TRANSLATION`? the `START `stage will only be used when a translation is missing a translator (this can happen if a translator drops a story translation). The naming is confusing and we should think about better stage names.

backend/python/app/models/story_translation.py
ZhangCreations: SQLAlchemy and Graphene's enum types aren't implicity castable so trying to use an enum here and in SQLAlchemy makes the code not work. I saw this approached used on the stories model (albeit for slightly different reasons) so I'm not 100% sure what the call was here, forgot to include this in the intial pr sorry :/

gaoxk: hmm interesting, enums are pretty wack. It might get even weirder when we switch to mysql so keeping this as a string for now is fine too. ty for explaining :)

gaoxk (approved):
🚢

PR #25, opened 2021-06-12
Fix bug in content loading for story queries

e-wai:
## Notion ticket link

[Fully load content on storyByID](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=d2730d8328e948b4be291b19c2aaf126)

## Implementation description
* I don't know why this fixes anything I'm sorry

## Steps to test
1. Set up your DBs, make sure that you have at least one story
2. Test query:
```
{
  storyById(id: 2) {
    id
    title
    contents {
      id
      storyId
      lineIndex
      content
    }
  }
  stories {
    id
    contents {
      id
      storyId
      lineIndex
      content
    }
  }
}
```
3. Make sure `contents` field is not null on either! Either empty array `"contents": []` or something like
```
"contents": [
  {
    "id": 1,
    "storyId": "1",
    "lineIndex": 0,
    "content": "Line 1"
  },
  {
    "id": 2,
    "storyId": "1",
    "lineIndex": 1,
    "content": "Line 2"
  }
]
```

* _why does this work_ 🐋 🐳 🐋 
* About lazy-loading, I chose default behaviour there are sufficient cases where I see us querying without wanting all the content details. Reference here -> https://docs-sqlalchemy.readthedocs.io/ko/latest/orm/loading_relationships.html

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -13,7 +13,9 @@ class StoryService(IStoryService):
     def get_stories(self):
         # Entity is a SQLAlchemy model, we can use convenient methods provided
         # by SQLAlchemy like query.all() to query the data
-        return [result.to_dict() for result in Story.query.all()]
+        return [
+            story.to_dict(include_relationships=True) for story in Story.query.all()
+        ]
 
     def get_story(self, id):
         # get queries by the primary key, which is id for the Story table
@@ -21,7 +23,7 @@ class StoryService(IStoryService):
         if story is None:
             self.logger.error("Invalid id")
             raise Exception("Invalid id")
-        return story.to_dict()
+        return story.to_dict(include_relationships=True)
 
     def create_story(self, story, content):
         # create story


gaoxk (changes requested):
Aha! Interesting, [here's ](https://github.com/uwblueprint/planet-read/blob/main/backend/python/app/models/story.py#L21) why `to_dict` was having trouble. We need to pass it` include_relationships=True` to get the contents. 
We should still err on the side of using `to_dict` to make sure we're never returning more information than necessary. Returning the model object directly runs that risk.

e-wai:
Hmm it's still not working when I toggle `include_relationships` to `True`

gaoxk:
🤔 are you sure? It works locally for me
changed lines are
```
return [result.to_dict(include_relationships=True) for result in Story.query.all()]
...
return story.to_dict(include_relationships=True)
```

e-wai:
Ah, I was making changes to StoryContent instead of Story. Should be fixed!

gaoxk (approved):
:shipit:

PR #27, opened 2021-06-13
Homepage/list stories to translate

lynn-kim:
## Notion ticket link

[List all stories available to be translated by User](https://www.notion.so/uwblueprintexecs/List-all-stories-available-to-be-translated-by-User-7f85a3688844460d8c79dc8dcdb0417d)

## Implementation description
* add query for if a story is available for translation to a user 
* checks for stories that are not already translated in the user's approved languages, and that the level of the story is equal to or less than the approved language level 

## Steps to test
1. Seed DB with some users with varying language level pair 

example: 
```
docker exec -it <db container id> psql -U postgres -d planet-read
INSERT INTO users (first_name, last_name, auth_id, role, approved_languages) VALUES ('First', 'Last', 'insert-firebase-uid', 'Admin', '{"english":3, "french": 1}');
```
2. Seed DB with stories 
 ```
INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('my story title', 'story description', 'link', 1, '{"hindi", "french"});
```

* make sure that the same story is not returned more than once (each story is unique)
* test that the query filters are correct (there is at least one approved language that is not in the story's translated languages AND the language level is <= approved language) 
* `get_stories_available_for_translation` in `app/services/implementations/story_service.py`

## Checklist

### CODE REVIEW

File: backend/python/app/graphql/schema.py [hunk 1]
```diff
@@ -38,6 +39,9 @@ class Query(graphene.ObjectType):
     users = graphene.Field(graphene.List(UserDTO))
     user_by_id = graphene.Field(UserDTO, id=graphene.Int())
     user_by_email = graphene.Field(UserDTO, email=graphene.String())
+    stories_available_for_translation = graphene.Field(
+        graphene.List(StoryResponseDTO), id=graphene.Int()
```
gaoxk, line 43: nit: `id`->`user_id` to be crystal clear

backend/python/app/services/implementations/story_service.py
```diff
@@ -55,6 +55,23 @@ def create_story(self, story, content):
...
+        approved_languages = user.approved_languages
+        for lang in approved_languages:
+            level = approved_languages[lang]
+            stories = (
+                Story.query.filter(Story.level <= level)
+                .filter(~Story.translated_languages.any(lang))
+                .all()
+            )
+
+            for story in stories:
```
gaoxk, line 69: If the below `if `statement that I commented on works, I think we can skip this loop in favour of making `all_stories` a set and union-ing `stories` with it.
https://www.w3schools.com/python/ref_set_union.asp

backend/python/app/services/implementations/story_service.py
```diff
@@ -55,6 +55,23 @@ def create_story(self, story, content):
...
+                if not story in all_stories:
```
gaoxk, line 70: wait - does this if statement work? 😳 story is a story model and each model would have its own pointer

backend/python/app/services/implementations/story_service.py
```diff
@@ -55,6 +55,23 @@ def create_story(self, story, content):
...
+                    all_stories.append(story)
+
+        return all_stories
```
gaoxk, line 73: Let's do something like
`return [story.to_dict(include_relationships=True) for story in all_stories]`
Its a good habit to return with to_dict incase models have any sensitive info

gaoxk (changes requested):

backend/python/app/services/implementations/story_service.py
```diff
@@ -55,6 +55,21 @@ def create_story(self, story, content):
 
         return new_story
 
+    def get_stories_available_for_translation(self, user):
```
gaoxk, line 58: Found an error: 
```
stories 
id | title |       description       |                youtube_link                 | level | translated_languages
----+-------+-------------------------+---------------------------------------------+-------+----------------------
  5 | cill  | heres my storyy         | https://www.youtube.com/watch?v=5qap5aO4i9A |     1 | {ARABIC}
  6 | cill  | hindi story!!!          | https://www.youtube.com/watch?v=5qap5aO4i9A |     1 | {HINDI}
  7 | cill  | french story!!!         | https://www.youtube.com/watch?v=5qap5aO4i9A |     1 | {FRENCH}
  8 | cill  | french + hindi story!!! | https://www.youtube.com/watch?v=5qap5aO4i9A |     1 | {FRENCH,HINDI}
  
  users
  id | first_name  | last_name |           auth_id            | role  | resume | profile_pic |   approved_languages
----+-------------+-----------+------------------------------+-------+--------+-------------+-------------------------
  2 | PlanetAdmin | Read      | rHBoE0GuVFThaWgEUYj89IZ9QvZ2 | Admin |        |             | {"HINDI":3, "FRENCH":3}
```
`get_stories_available_for_translation` returns stories 5, 6, 7.

However, this reminds me that the language filter will always be set on the frontend. Let's have `get_stories_available_for_translation` accept a `language` rather than a `user`. Sorry that I didn't make this realization earlier 🤦

gaoxk (changes requested):

backend/python/app/services/implementations/story_service.py
gaoxk: You'll actually still need the `user` to an extent - you'll still have to look at the approved_languages on the `user` to make the appropriate level filter.

backend/python/app/services/implementations/story_service.py
```diff
@@ -57,6 +57,14 @@ def create_story(self, story, content):
 
         return new_story
 
+    def get_stories_available_for_translation(self, language, level):
+        stories = set(
```
gaoxk, line 61: Do we need a `set` here? `Story.query` should never return duplicate stories unless some weird joining goes on

gaoxk (approved):
🚢 🚢 🚢 🚢 🚢  nice! thanks for being flexible.

PR #31, opened 2021-06-16
Change ResetPassword props declaration from type to interface

e-wai:
## Notion ticket link
Dev health thing

## Implementation description
* Change from interface to type declaration

## Steps to test
1. Start localhost and try reset password when logged in and logged out (with real / not real input)

* More of me being unsure about Typescript conventions. This seems to imply that default props that I'm relying on will be deprecated (https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/default_props/, https://github.com/reactjs/rfcs/pull/107). To be replaced by object default values but we're using functional components everywhere?
* Otherwise, we could declare the type as `string | null` and throw some nullchecks in?

## Checklist

### DIFF
--- a/frontend/src/components/auth/ResetPassword.tsx
+++ b/frontend/src/components/auth/ResetPassword.tsx
@@ -2,9 +2,9 @@ import React, { useContext } from "react";
 import { gql, useMutation } from "@apollo/client";
 import AuthContext from "../../contexts/AuthContext";
 
-interface ResetPasswordProps {
+type ResetPasswordProps = {
   email?: string;
-}
+};
 
 const RESET_PASSWORD = gql`
   mutation ResetPassword($email: String!) {
--- a/frontend/src/components/auth/ResetPassword.tsx
+++ b/frontend/src/components/auth/ResetPassword.tsx
@@ -1,9 +1,8 @@
-import React, { useContext } from "react";
+import React from "react";
 import { gql, useMutation } from "@apollo/client";
-import AuthContext from "../../contexts/AuthContext";
 
 type ResetPasswordProps = {
-  email?: string;
+  email: string;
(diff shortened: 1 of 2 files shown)


gaoxk (changes requested):
Regarding default props, what you're doing won't be deprecated. `ResetPaswordProps` is defining the type of the props, not the default values of the props - you're in the clear! Declaring defaultprops for the resetpassword component would look like
```
const defaultProps = {
  email : "doit@forthe.vine",
};
....
ResetPassword.defaultProps = defaultProps;
```
Regarding `email : string | null` vs `email?: string`, the first indicates that email may be `null`, the second indicates that email might be `undefined`. [Here's](https://basarat.gitbook.io/typescript/recap/null-undefined) a good article reading up on the differences, namely, 
```
Something hasn't been initialized : undefined.
Something is currently unavailable: null.
```
Now that I think about it, we should require `email` to be initialized. What happens if you remove the `?`?

gaoxk:
(you dont have to change it to `?` if it doesn't work out. I'm requesting changes moreso as a call to action on investigating this.)

e-wai:
So I still think I use default props here: https://github.com/uwblueprint/planet-read/blob/8657f89241ff28de6b2bf3884441dbf62d488153/frontend/src/components/auth/ResetPassword.tsx#L69-L71

The reason that the parameter was chosen to be optional is for the case that the user is resetting password from the default screen while logged in - in which case the email string is from the `authenticatedUser` context rather than from the log in screen text box. I can take out the `?` making it optional if I pass the email instead as a prop (referencing context) from the `Default.tsx` page. Maybe that makes more sense?

gaoxk (approved):
🚢

PR #32, opened 2021-06-17
Get Stories Available for Review by User

andrewyguo:
## Notion ticket link

[List all stories available to be reviewed by User](https://www.notion.so/uwblueprintexecs/List-all-stories-available-to-be-reviewed-by-User-40702ca6c0ef4daab38a9e2d997aed03)

## Implementation description
*  Created an additional story query that retrieves all of the stories available to be reviewed
*  Given a level and a language, query will return all story translations in that language that can be reviewed by someone with these credentials 

## Steps to test
1. Seed database with different stories and seed database with different story_translations that have a reviewer_id of None 
Example: 
`INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('Kevin Eats Caviar for Lunch', 'He complains its too cheap', 'link', 4, '{"GERMAN", "ENGLISH_UK"}');`
`INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('Kevin Eats Gold', 'He complains its too cheap', 'link', 2, '{"GERMAN", "MANDARIN"}');`
Insert into StoryTranslations: 
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (1,  'FRENCH', 'REVIEW', 1);`
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (2,  'ARABIC', 'REVIEW', 1);`
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (3,  'ARABIC', 'REVIEW', 1);`
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (2,  'GERMAN', 'REVIEW', 1);`
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (1,  'ENGLISH_US', 'START', 1);`
`INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (1,  'RUSSIAN', 'PUBLISH', 1);`

2. Call the query with various levels and languages and see returned list of story_translations

## What should reviewers focus on?
* Ensuring that the query returns story_translations as expected

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] For backend changes, I have run the appropriate linters:  `docker exec -it <backend-container-id> /bin/bash -c "black . && isort --profile black ."`
[...]

PR #35, opened 2021-06-21
Establish homepage in default state

e-wai:
## Notion ticket link

[Create homepage and state](https://www.notion.so/uwblueprintexecs/Create-homepage-and-state-34433422ca004acda806a888f2711f92)

## Implementation description
* Created `Header` component with CSS. No interaction right now -> added logo and placeholder user icon assets into new `assets` directory as SVGs
* Created `ToggleButton` component with CSS. Basically a fancy checkbox, but there's a point of discussion below. Referenced https://webomnizz.com/how-to-create-a-toggle-switch-button-in-react/, not sure if I'm using all of the many `divs` and `spans` correctly
* Implemented query logic in `HomePage` component. `Enum` file in constants directory for general purpose

Changes to `StoryCard` and `StoryList`:
* added CSS (and some was brought out to parent component `HomePage`)
* changed Props so that it also includes `id/storyId` (same concept, different name depending on the table queried so aliased to `storyId` in https://github.com/uwblueprint/planet-read/blob/69ad057f2da02df2f2471cb200376ae9a6e15e08/frontend/src/components/pages/HomePage.tsx#L42) and optionally `storyTranslationId` if it's a translation. This was discussed with @jennchenn and @gaoxk on Slack. [Update: this PR was rebased on top of the recently-merged https://github.com/uwblueprint/planet-read/pull/36. Should consider everything that's happening]
* `cardId` created as a key for mapping in `StoryList` out of `storyId` and `storyTranslationId` if it exists (this forces uniqueness, `storyId` will only show up once without `storyTranslationId` and may show up multiple times with different `storyTranslationId`s. This `cardId` is not used anywhere else. 

## Steps to test
1. Set up DBs with some good test data. At some point, try with one of the queries returning an empty array to see the <NoStoryCard /> pop up. This is what my `story_translation_contents` table looked like, mostly the test data in https://github.com/uwblueprint/planet-read/pull/32 with the last entry changed to have a `reviewer_id`:
image
2. Run everything and hop on over to `/stories`. If you're using the test data above, please log in as whichever is user 1 for you. 

For 3 and 4:
homepage default state done

3. On the first visit, this should load the "My work" section with your translation stories. Check that the expected story cards are rendered.
4. Toggle over to the "Browse stories" section. Check that the expected story cards are rendered.
5. Change your role to reviewer. This has to be done manually: https://github.com/uwblueprint/planet-read/pull/35/files#r656720490 Check "My work" and "Browse stories" again.

* Left a comment in `Header.css`. You can see the scroll bar pop up and shift everything in the first gif commented below, what's the expected behaviour?
> this width may change on toggle
> This happens when the cursor bar on the right is asked to appear
> aka when My Stories has # of entries that > screen and Browse doesn't
> or vice versa

* Left a comment in `ToggleButton.tsx`. I tried to make the component re-usable, so it doesn't know details like the option names, however it _does_ default to true on the left always. Maybe this is a UX question, but is that ok?
>    How can we make it super visible that the true state is always displayed on the left
>    Or do we want this to change?
>    Right now, just specified in logic / className
>    
>    Left = checked = true
>    Right = unchecked = false

* Is anything taking on more responsibility than they should? (ie. more stateful than it should be)

* Is there any instances where I misuse `id` vs. `className` attributes? 
* There's just no error handling :yikes: -> how should I be handling errors/what errors happen?
* So much `px` size hardcoding, that's perfectly right, right?

## Checklist


frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,31 +1,172 @@
...
+  MyWork,
+  Browse,
+}
+
+type QueryInformation = {
+  fieldName: string;
+  string: DocumentNode;
+};
+
+type StoryResponseDTO = Omit<StoryCardProps, "cardId"> & {
```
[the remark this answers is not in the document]
e-wai: I also don't know if the alias is necessary or if the mapping works out fine w/o it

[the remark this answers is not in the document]
e-wai: Huh ok I tried something new. I think most of the above can be ignored. Now,
- Everything from the GraphQL call is passed into the `StoryCardProps`, including the potentially-DNE `storyTranslationId`. `cardId` is introduced in `StoryList` to create a key. Thoughts here:
https://github.com/uwblueprint/planet-read/blob/6e96df77fffd686988a1c690c0b5a2b9450a6725/frontend/src/components/story/StoryList.tsx#L16-L22

e-wai:
@jennchenn @ArmanyaDalmia 

Still a bunch of cleaning up to do, but barring disaster, should be a pretty decent look at this PR's structure? Let me know if I'm doing anything dumb or if you have any questions

homepage default toggle

frontend/src/components/navigation/Header.css
```diff
@@ -0,0 +1,32 @@
...
+    padding: 20px 0px 20px 0px;
+}
+
+.header-information {
+    align-items: center;
+    background-color: #fafafa;
+    display: flex;
+    flex-direction: row;
+    justify-content: left;
+    /* TODO: this width changes on toggle?? */
```
gaoxk, line 15: Looks like the `homepage-display-state-toggle-section` is also changing width on toggle 😳

frontend/src/components/navigation/ToggleButton.css
```diff
@@ -0,0 +1,40 @@
...
+    min-width: 150px;
+    padding: 5px 40px 5px 30px;
+}
+
+.selected {
+    background-color: #333333;
+}
+
+.selected span {
+    /* TODO caps or lowercase */
```
gaoxk, line 30: caps 😎

frontend/src/components/pages/HomePage.css
```diff
@@ -0,0 +1,8 @@
+#homepage-display-state-toggle-section {
+    background-color: #FAFAFA;
+    padding-top: 5px;
+    padding-bottom: 5px;
+    min-width: 800px;
+    width: 60%;
+    /* TODO who should manage the width of story cards - should be same width as this section */
```
gaoxk, line 7: components take on the width of their parent component by default. To prevent any inconsistency, you can wrap both the header and the list of story cards in the same div, and set the width of the div manually (eg `width: 60%;`)

frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,31 +1,172 @@
...
+  // cardId will be assembled based on these fields
+  storyId: number;
+  storyTranslationId?: number;
+};
 
 const HomePage = () => {
+  // how do private methods exist?
+  const buildHomePageStoriesQuery = (
```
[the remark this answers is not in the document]
gaoxk: > There's a bit of repetition that I kept in for clarity. It makes it very clear what the query is. Thoughts?

It seems that theres only 3 lines changing in the gql query strings themselves, eg:
```
storyTranslationsByUser(
            userId: ${userId},
            translator: ${role === Role.Translator}
          )
```
Perhaps the rest of the query string can be abstracted away and the returns in this function can be more succinct

>WHy does else if not work??? The linter doesn't like it. Should I be using switch-cases?

Is the error you're getting `Unnecessary 'else' after 'return'  no-else-return`? To avoid this, you can initialize a return variable and change it in the if statement
```
let res = {fieldName:"storyTranslationsByUser", string:... }
if (...){
  res = ....
} else if (...) {
  res = ....
}
return res;
```
> Where should these private-type methods exist? As in not used outside of component, not re-used -> it's only in its own method to clean up the reading for the bottom lines.

two thoughts: 
- Building query strings for repeat use is actually desirable - it encourages devs to have consistent expectations of DTO shape. It can also help us out in funky situations like the `refresh` mutation, where the user **must** request for a refresh_token. Something we should look into is storing and managing all our queries
- What you have going on here indeed scopes the function to `HomePage` - which a 'private function' in this setting. More about private functions in prototypes and classes [here](https://stackoverflow.com/questions/55611/javascript-private-methods?page=1&tab=votes#tab-top) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields). Note that we're likely to never explicitly write a prototype or class in this codebase - react is real f u n c t i o n a l these days.

frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,31 +1,133 @@
...
+  const [role, setRole] = useState<Role>(Role.Translator);
+  const [maxLevel, setMaxLevel] = useState<number>(1);
   const [stories, setStories] = useState<StoryCardProps[] | null>(null);
 
-  useQuery(GET_STORIES, {
+  const { authenticatedUser } = useContext(AuthContext);
+
+  // TODO: dummy setting to avoid the unused vars error :(
+  if (authenticatedUser == null) {
+    setDisplayMyStories(true);
```
gaoxk, line 98: to avoid linting issues for currently unused state setters, I believe you can skip declaring them for now + remove this if statement. ie:
`const [role] = useState<Role>(Role.Translator);`

frontend/src/components/story/StoryList.tsx
```diff
@@ -5,17 +5,52 @@ export type StoryListProps = {
   stories: StoryCardProps[] | null;
 };
 
+const NoStoriesFoundCard = () => {
+  // TODO: is this a re-usable component? Deserve its own class?
```
gaoxk, line 9: I think NoStoriesFoundCard is fine here

frontend/src/components/story/StoryList.tsx
```diff
@@ -5,17 +5,52 @@ export type StoryListProps = {
...
-      key={`story-card-key-${story.id}`}
-      id={story.id}
-      title={story.title}
-      description={story.description}
-      youtubeLink={story.youtubeLink}
-      level={story.level}
-    />
-  ));
+  /**
+   * I think this will recalculate everytime we re-render. Is this avoidable? Is having better handled in HomePage?
```
gaoxk, line 19: I think this operation is arbitrary enough that it isn't a concern. 
Here's a snazzy article if you're thinking about using indexes instead: ["Index as a key is an anti pattern"](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318)

gaoxk (commented):
👯 It's really coming along!

frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,32 +1,123 @@
...
+        `,
+      };
+    }
+
+    return result as QueryInformation;
+  };
+
+  const [displayMyStories, setDisplayMyStories] = useState<boolean>(true);
+  const [translationLanguage] = useState<string>("RUSSIAN");
+  const [role] = useState<Role>(Role.Translator);
```
e-wai, line 87:
```suggestion
  const [role] = useState<Role>(Role.Reviewer);
```

e-wai: For testing description

frontend/src/components/story/StoryCard.tsx
```diff
@@ -30,8 +32,12 @@ const StoryCard = ({
     return originalYoutubeLink.replace("watch?v=", "embed/");
   };
 
+  // for on-click. StoryTranslationId may be undefined?
```
gaoxk, line 35: nit: remove console logs

frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,32 +1,123 @@
...
+    role,
+    maxLevel,
+    +authenticatedUser!!.id,
+  );
+
+  useQuery(query.string, {
     fetchPolicy: "cache-and-network",
-    onCompleted: (data) => setStories(data.stories),
+    onCompleted: (data) => setStories(data[query.fieldName]),
+    // Assumes independent queries (query will never include two sub-queries)
```
gaoxk, line 104: nit: move comment above `onCompleted` line

frontend/src/components/navigation/ToggleButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const ToggleButton = ({
+  currentState,
+  trueStateName,
+  falseStateName,
+  onToggle,
+}: ToggleButtonProps) => {
+  /**
+   * How can we make it super visible that the true state is always displayed on the left
```
gaoxk, line 18: Rather than true vs false (state) we can think of primary vs secondary (state), where left is default selected. left vs right wording may also be clear. Open to thoughts on naming. 

`trueStateName` does feel like an odd name for a toggle button label. What do you think of `primaryToggleName`?

gaoxk (changes requested):

frontend/src/components/story/StoryCard.tsx
e-wai: I needed it to pass no-unused-vars error

frontend/src/components/navigation/ToggleButton.tsx
e-wai: how about `LeftToggleLabel`?

e-wai:
Post-rebase state 🎉 🎉 🎉 
homepage default loading and card buttons

frontend/src/components/navigation/ToggleButton.tsx
gaoxk: `LeftToggleLabel` sounds good

gaoxk (approved):
Nice!

PR #39, opened 2021-07-01
Create TranslationPage for translating story content

e-wai:
## Notion ticket link

[Load all story contents into cells/table](https://www.notion.so/uwblueprintexecs/Load-all-story-contents-into-cells-table-9d7c0ffb11954bb4a05c85b18920d717)

## Implementation description
(Commits should make it easier to review code and pieces and is an ok description of implementation:)
* Add components for page: image icons, `Cell` (story content), `EditableCell` (story translation content)
* Create `TranslationPage` to house these components. About how to store the story content / translation content -> array was the easier choice as it's ordered, maps easily, and passing content id / line Index around requests meant you could set quickly using indices (without looping to find). Back half of the `StoryLine` object is optional as translated content set in a different loop from story content.
* There's a couple bugs around story translation content queries/mutations, this fixes one of them (that it wasn't saving the changes into db when called)

## Steps to test
1. From the stories page, go to a translations route -> pressing `view translation / edit translation / translate book` should lead to a `/translation/storyId/storyTranslationId` route ... some of these pages are broken
2. Edit text boxes to make the translation. You can only do this slowly - the changes don't register properly when you do it quickly. However, you can change it not one letter at a time by copy/pasting longer messages. Note that the icons flicker (they are appear randomly, without any logic yet).
3. You can check that the changes are being saved properly: the easiest way is to refresh the page and ensure that the changes made still appear (and are thus saved somewhere). You can also open up the db and check that changes are being registered properly.
translation load content

`<TextArea>` doesn't grow automatically, but reportedly `<Input>` doesn't do multi-line 

## What should reviewers focus on?
* I stayed fairly light on the CSS. Do we have finalized designs or is this mostly placeholder until the stylized components to finish up?
* More CSS: I used flex for rows instead of grid... is that a bad decision from a best practices / efficiency standpoint?
* Is there a preferred name for the translation page route?
* (I think we talked about this, but just noting -> this re-renders the entire page for every change and makes a new query every single time. That seems less than ideal)
* How do the dimensions look on your browser? 
* Should create a ticket to investigate and double check which queries/mutations are broken? I'm seeing some null errors with `StoryTranslationsAvailableForReview` and `StoryTranslationsByUser` 
* Are there error states to cover, and do we want a loading card?

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
[...]

PR #40, opened 2021-07-02
Add translation progress bar

jennchenn:
## Notion ticket link
[Add translation progress state](https://www.notion.so/uwblueprintexecs/Add-translation-progress-state-d384e35a5f3e49bf89d3ed64d7e52573)

## Implementation description
* Added progress bar to TranslationPage in frontend
* Counting and returning number of translated lines on `storyTranslationById` query (added `numTranslatedLines` to `StoryTranslationResponseDTO`)
* Will be building implementation in frontend to reflect/update percentage progress bar in frontend after PR #39 is merged

## Steps to test
1. Seed database with data:
```
$ INSERT INTO users (first_name, last_name, auth_id, role, approved_languages) VALUES ('PlanetAdmin', 'Read', <auth id>, <role>, '{"HINDI":3}');
$ INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('A Book for Puchku', 'Puchku has run out of books to read. Then she discovers more books in the top shelf of the library bookcase. But Puchku is small, and the bookcase tall. How will she ever get to her beloved books?', 'https://www.youtube.com/watch?v=PKxMjO4HPvE', 3, '{}');
$ INSERT INTO story_contents (story_id, line_index, content) VALUES (<id of story you just inserted>,1,'first line'), (<id of story you just inserted>,2,'second line'), (<id of story you just inserted>,3,'third line'), (<id of story you just inserted>,4,'fourth line');
$ INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (<id of story>,'HINDI', 'TRANSLATE',1);
$ INSERT INTO story_translation_contents (story_translation_id, line_index, translation_content) VALUES (<id of story_translation>, 1, 'translation!');
$ INSERT INTO story_translation_contents (story_translation_id, line_index, translation_content) VALUES (<id of story_translation>, 2, '');
$ INSERT INTO story_translation_contents (story_translation_id, line_index, translation_content) VALUES (<id of story_translation>, 3, '');
```
2. Check that query for storyTranslationById returns correct number of translated lines (i.e. in this case it should return 1):
```
{
  storyTranslationById(id: 1) {
    id
    numTranslatedLines
  }
}
```
3. Verify the CSS for the progress bar when you navigate to `http://localhost:3000/translation/{storyId}/{storyTranslationId}` (currently hardcoded at 25%)
image

* `backend/python/app/services/implementations/story_service.py`
* `frontend/src/components/translation/TranslationProgressBar.tsx`

## Checklist
- [x] For backend changes, I have run the appropriate linters:  `docker exec -it planet-read_py-backend_1 /bin/bash -c "black . && isort --profile black ."`

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -2,7 +2,6 @@ from flask import current_app
 
 from ...graphql.types.story_type import (
     StoryTranslationContentResponseDTO,
-    StoryTranslationProgressResponseDTO,
 )
 from ...models import db
 from ...models.story import Story
@@ -131,22 +130,29 @@ class StoryService(IStoryService):
             # TODO: Condense into singular query
             translation = StoryTranslation.query.get(id)
 
+            # story_details = Story.query.
             story_details = (
-                db.session.query(
-                    Story.id.label("story_id"),
-                    Story.title.label("title"),
-                    Story.description.label("description"),
-                    Story.youtube_link.label("youtube_link"),
-                    Story.level.label("level"),
+                (
+                    db.session.query(
+                        Story.id.label("story_id"),
+                        Story.title.label("title"),
+                        Story.description.label("description"),
+                        Story.youtube_link.label("youtube_link"),
+                        Story.level.label("level"),
+                    )
+                    .join(StoryTranslation, Story.id == StoryTranslation.story_id)
+                    .filter(StoryTranslation.id == id)
                 )
-                .join(StoryTranslation, Story.id == StoryTranslation.story_id)
-                .filter(StoryTranslation.id == id)
                 .one()
+                ._asdict()
             )
 
+            story_details["percentage_complete"] = self._get_translation_progress(
+                story_details["story_id"], id
+            )
             response = {
                 **translation.to_dict(include_relationships=True),
-                **story_details._asdict(),
+                **story_details,
             }
             return response
 
@@ -174,8 +180,19 @@ class StoryService(IStoryService):
 
     def update_story_translation_content(self, story_translation_content):
         try:
-            old_translation_content = StoryTranslationContent.query.get(
-                story_translation_content.id
+            old_translation_content = (
+                db.session.query(
+                    StoryTranslation.id,
+                    StoryTranslation.story_id,
+                    StoryTranslationContent.translation_content,
+                    StoryTranslationContent.line_index,
+                )
+                .join(
+                    StoryTranslation,
+                    StoryTranslation.id == StoryTranslationContent.story_translation_id,
+                )
+                .filter(StoryTranslationContent.id == story_translation_content.id)
+                .one()
             )
 
             if not old_translation_content:
@@ -189,7 +206,15 @@ class StoryService(IStoryService):
                 id=story_translation_content.id
             ).first()
 
-            story_translation.translation_content: story_translation_content.translation_content
+            story_translation.translation_content = (
+                story_translation_content.translation_content
+            )
+            db.session.commit()
+
+            percentage_complete = self._get_translation_progress(
+                old_translation_content.story_id, old_translation_content.id
+            )
+
             db.session.commit()
         except Exception as error:
             reason = getattr(error, "message", None)
@@ -201,9 +226,10 @@ class StoryService(IStoryService):
             raise error
 
         return StoryTranslationContentResponseDTO(
-            id=story_translation_content.id,
-            line_index=old_translation_content.line_index,
-            translation_content=story_translation_content.translation_content,
+            story_translation_content.id,
+            old_translation_content.line_index,
+            story_translation_content.translation_content,
+            percentage_complete,
         )
 
     def update_story_translation_contents(self, story_translation_contents):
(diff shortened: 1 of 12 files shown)


frontend/src/components/pages/TranslationPage.css
```diff
@@ -0,0 +1,16 @@
+.translation {
```
gaoxk, line 1: very nice looking progress bar 👌 did you run it by Shaahana?

gaoxk (changes requested):

backend/python/app/services/implementations/story_service.py
```diff
@@ -240,3 +243,22 @@ def get_story_translations_available_for_review(self, language, level):
...
+                db.session.query(StoryTranslationContent)
+                .filter(StoryTranslationContent.story_translation_id == translation_id)
+                .count()
+            )
+            num_story_lines = (
+                db.session.query(StoryContent)
+                .filter(StoryContent.story_id == story_id)
+                .count()
+            )
+            percentage_complete = round(num_translated_lines / num_story_lines, 2) * 100
```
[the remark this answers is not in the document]
jennchenn: ahh sorry i had meant to add in a check for non-empty strings as well like this, but i accidentally deleted it when i added in the count():
```
num_translated_lines = (
                db.session.query(StoryTranslationContent)
                .filter(
                    StoryTranslationContent.story_translation_id == translation_id,
                    StoryTranslationContent.translation_content != "",
                )
                .count()
            )
```
would that work to return the correct percentage?

i can definitely move it into `get_story_translation`--would the translationContent that gets returned in `StoryTranslationResponseDTO` only include the non-empty string values? also how often do we want the progress bar to be updated, and do we need to call get_story_translation multiple times on the translation page as well?

frontend/src/components/pages/TranslationPage.css
jennchenn: sorry completely slipped my mind just messaged i will make any updates according to her comments

backend/python/app/services/implementations/story_service.py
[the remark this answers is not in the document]
e-wai: Heads up that
> alternative idea: modify StoryTranslationResponseDTO to return translationContent (@e-wai already has this in her WIP pr #39 )

is currently not working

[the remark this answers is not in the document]
gaoxk: >[content in dto] is currently not working

@e-wai  We can extend this conversation on #39 but my guess is that StoryTranslation model needs something like the `content` SQLAlchemy attribute in the Story model

> would the translationContent that gets returned in StoryTranslationResponseDTO only include the non-empty string values?

If we follow the above suggestion, the query would return all content lines regardless of actual string content. 

> also how often do we want the progress bar to be updated, and do we need to call get_story_translation multiple times on the translation page as well?

We're taking a [direct to database approach (stackoverflow ans)](https://stackoverflow.com/questions/44435034/auto-save-server-architecture) for now to auto save translation changes. So we will certainly be making alot of update mutations - I think we should write these updates to return the storytranslation + it's content, and therefore we can return the `percentageComplete` too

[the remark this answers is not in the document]
e-wai: Do we want to store `percentageComplete` as a column on StoryTranslation?
- on update (mutation) -> recalculate. However, if we were to add as a column, we would have be editing two tables (is this expensive?)
- on query for story translation -> if it's stored as a column, this isn't a problem and can just a direct read

Mini optimization frontend -> having logic to check existing string value, "" vs. "abc". If it's the latter, don't even bother with the recalculate portion of the call since any changes won't affect progress.
- However, there's an edge case where if the user is going from "blah" to "", should we be toggling progress back?

backend/python/app/services/implementations/story_service.py
```diff
@@ -245,3 +249,6 @@ def get_story_translations_available_for_review(self, language, level):
         except Exception as error:
             self.logger.error(str(error))
             raise error
+
+    def _get_num_translated_lines(self, translation_contents):
+        return sum(1 for _ in translation_contents if _.translation_content)
```
e-wai, line 254: How does this handle empty strings?

jennchenn: i believe Python empty strings evaluate to False so it should only sum if the translation_content is not an empty string--this seemed to be the case when i was testing but let me know if it's not working for you/if there's a more elegant way to do the check i'm happy to change it!!

frontend/src/components/translation/TranslationProgressBar.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { ProgressBar } from "react-bootstrap";
```
gaoxk, line 2: we haven't made any formal component library decisions so I'm happy with keeping the bootstrap. I am curious, do you have a particular reason for bootstrap or are you comfortable with it?

gaoxk (changes requested):
Nice! Just left some quick comments

frontend/src/App.tsx
```diff
@@ -35,6 +36,11 @@ const App = () => {
           <PrivateRoute exact path="/entity/update" component={UpdatePage} />
           <PrivateRoute exact path="/entity" component={DisplayPage} />
           <PrivateRoute exact path="/stories" component={HomePage} />
+          <PrivateRoute
```
gaoxk, line 39: heads up: may need rebase if jenn merges first

frontend/src/components/translation/TranslationProgressBar.tsx
jennchenn: oops i can definitely swap it out later once we decide on a component library! not too much thought behind it, but thought it would be easier to quickly spin up a functional progress bar
i thought i had saw bootstrap used elsewhere in the repo but i think i just imagined it now 😅
i have used it a few times before but definitely can switch it to something else!!

backend/python/app/services/implementations/story_service.py
```diff
@@ -129,21 +129,27 @@ def get_story_translation(self, id):
             translation = StoryTranslation.query.get(id)
 
             story_details = (
-                db.session.query(
-                    Story.id.label("story_id"),
-                    Story.title.label("title"),
-                    Story.description.label("description"),
-                    Story.youtube_link.label("youtube_link"),
-                    Story.level.label("level"),
+                (
```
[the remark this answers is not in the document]
jennchenn: nope removed now! i must have accidentally added it in with the rebasing/formatting

gaoxk (approved):
🚢 🚢

frontend/src/App.tsx
e-wai: Will rebase 👍

PR #41, opened 2021-07-06
Modify query to return story translation contents in response

e-wai:
## Notion ticket link

Some unblocking work for https://github.com/uwblueprint/planet-read/pull/39 and https://github.com/uwblueprint/planet-read/pull/40

## Implementation description
- Query the `StoryTranslation` table, including the include_relationships trait for the conversion to dictionary so that it includes the `translation_content` field
- Do a session query on the `Story` table, joining with the `StoryTranslation` table to know which story to look at, and convert to a dictionary object
- Combine these two dictionary objects into one response dictionary
- Rename the `StoryTranslation` table's id attribute to `story_translation_id`

## Steps to test
1. Try out all the queries and mutations around `StoryTranslationResponseDTO`. Check all the fields are there (my go-to is checking something on the Story table, StoryTranslation table, and translationContent)
- Query: `storyTranslationById`, `storyTranslationsByUser`, `storyTranslationsAvailableForReview`
- Mutation: `updateStoryTranslationContentById`, `updateStoryTranslationContents`

* See comment

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -125,23 +125,28 @@ class StoryService(IStoryService):
 
     def get_story_translation(self, id):
         try:
-            return (
+            # TODO: Condense into singular query
+            translation = StoryTranslation.query.get(id)
+
+            story_details = (
                 db.session.query(
                     Story.id.label("story_id"),
                     Story.title.label("title"),
                     Story.description.label("description"),
                     Story.youtube_link.label("youtube_link"),
                     Story.level.label("level"),
-                    StoryTranslation.id.label("story_translation_id"),
-                    StoryTranslation.language.label("language"),
-                    StoryTranslation.stage.label("stage"),
-                    StoryTranslation.translator_id.label("translator_id"),
-                    StoryTranslation.reviewer_id.label("reviewer_id"),
                 )
                 .join(StoryTranslation, Story.id == StoryTranslation.story_id)
                 .filter(StoryTranslation.id == id)
                 .one()
             )
+
+            response = {
+                **translation.to_dict(include_relationships=True),
+                **story_details._asdict(),
+            }
+            return response
+
         except Exception as error:
             self.logger.error(str(error))
             raise error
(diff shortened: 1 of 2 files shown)


backend/python/app/services/implementations/story_service.py
```diff
@@ -125,23 +125,28 @@ def get_story_translations(self, user_id, translator, language, level):
...
+            translation = StoryTranslation.query.get(id)
+
+            story_details = (
                 db.session.query(
                     Story.id.label("story_id"),
                     Story.title.label("title"),
                     Story.description.label("description"),
                     Story.youtube_link.label("youtube_link"),
                     Story.level.label("level"),
-                    StoryTranslation.id.label("story_translation_id"),
```
e-wai, line 135: Should this be called `id` or `story_translation_id`? For consistency, it looks like id, but it seems less clear than the alternative

e-wai:
Right now working -> storyTranslationById, updateStoryTranslationContents

Don't work -> storyTranslationsByUser (translationContent null), storyTranslationsAvailableForReview (translationContent null), updateStoryTranslationContentById (trying to convert the str content to float?)

backend/python/app/services/implementations/story_service.py
gaoxk: now that i think about it, lets do `id`, consistency will help frontend development

backend/python/app/services/implementations/story_service.py
```diff
@@ -125,23 +125,28 @@ def get_story_translations(self, user_id, translator, language, level):
...
-                    StoryTranslation.stage.label("stage"),
-                    StoryTranslation.translator_id.label("translator_id"),
-                    StoryTranslation.reviewer_id.label("reviewer_id"),
                 )
                 .join(StoryTranslation, Story.id == StoryTranslation.story_id)
                 .filter(StoryTranslation.id == id)
                 .one()
             )
+
+            response = {
```
gaoxk, line 143: nice pattern! clean. 
Ideally, we would fetch all this info in 1 SQL call rather than 2, but I'm happy with this for now to unblock. Can you leave a TODO comment for reducing function to 1 SQL call for the future?

gaoxk (changes requested):

gaoxk (approved):

e-wai:
Outstanding ticket for remaining broken mutations to be created

PR #49, opened 2021-07-17
Add Chakra UI

gaoxk:
## Implementation description
* install chakra packages
* create theme objects with design system colours to pass to chakra provider
* convert Filter and StoryCard components to chakra ui components
  * moved css into theme except for non chakra ui component styling

## Steps to test
1.

* [thing about useStyleConfig pattern for Flex and Box]

## Checklist
- [ ] For backend changes, I have run the appropriate linters:  `docker exec -it <backend-container-id> /bin/bash -c "black . && isort --profile black ."`


frontend/src/components/homepage/Filter.tsx
```diff
@@ -40,43 +51,49 @@ const Filter = ({
       <option key={i.toString()} value={i.toString()} label={i.toString()} />,
     );
   }
+  const filterStyle = useStyleConfig("Filter");
```
gaoxk, line 54: `useStyleConfig` example

frontend/src/theme/colors.ts
```diff
@@ -0,0 +1,40 @@
+const colors = {
```
gaoxk, line 1: Not sure what color system is conducive to clear design<>dev mappings. `vividBlue` is more clear in naming but the `blue` object is  modifying the `blue` color scheme itself - i.e. `colorScheme="blue"` button props (`StoryCard.tsx`)
@ZhangCreations curious what our experience developing with colour was like in chakra ui

frontend/src/components/auth/Logout.tsx
```diff
@@ -32,7 +32,8 @@ const Logout = () => {
   useEffect(() => {
     if (error) {
       // TODO: Implement Error State
-      console.log(error);
+      // eslint-disable-next-line no-alert
```
gaoxk, line 35: linting errors

e-wai: Does your linter give you a blocking error when you don't include this?

frontend/src/theme/components/Filter.ts
```diff
@@ -0,0 +1,12 @@
+const Filter = {
+  baseStyle: {
+    backgroundColor: "#FAFAFA",
```
e-wai, line 3: see comment about colors being in `frontend/src/theme/colors.ts`? `#FAFAFA` is frequently used color

edit: OH FANCY IT SHOWS A COLOUR ICON

frontend/src/components/homepage/StoryList.css
```diff
@@ -1,17 +1,9 @@
 .no-stories-found-card, .loading-card {
     align-items: center;
     border: 2px solid #333333;
-    background-color: #FAFAFA;
+    background-color: #fafafa;
```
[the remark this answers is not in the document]
gaoxk: 👍 added  `backgroundSurfaceLoFi` to `theme/colors.ts`. The design system background is actually `#ECECEC` which looks much darker than `#FAFAFA` in action. We can delete `backgroundSurfaceLoFi` on the path to moving to hifi, as there are no final hifi designs.

frontend/src/components/homepage/Filter.tsx
```diff
@@ -25,8 +36,8 @@ const Filter = ({
   const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     setLanguage(event.target.value);
   };
-  const handleRangeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setLevel(parseInt(event.target.value, 10));
+  const handleRangeChange = (next: number) => {
+    setLevel(next);
   };
```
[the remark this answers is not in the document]
gaoxk: good point - on second thought, mentioning `level` makes more sense too

frontend/src/components/auth/Logout.tsx
gaoxk: the linter blocked me due to some weird local changes to prettier. I know this isn't technically blocking on master, but doesnt hurt to include it - `alerts` and `console.logs` are generally bad practice

frontend/src/components/pages/HomePage.tsx
```diff
@@ -4,7 +4,7 @@ import Filter from "../homepage/Filter";
 import StoryList from "../homepage/StoryList";
 import { StoryCardProps } from "../homepage/StoryCard";
 import AuthContext from "../../contexts/AuthContext";
-import ToggleButton from "../navigation/ToggleButton";
+import ToggleButton from "../utils/ToggleButton";
```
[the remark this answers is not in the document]
gaoxk: > Is there another name for "generic UI components" that we could throw ToggleButton into?

Good question! There is the [chakra ui switch](https://chakra-ui.com/docs/form/switch), but it's design needs a total overhaul
image

> Curious what other people think and maybe this is a normal practice I'm not aware of, but the double /utils folders with fairly different purposes is a bit trippy

I'm personally fine with it since the file is in `/components` in this case, but open to suggestions!

frontend/src/theme/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import colors from "./colors";
+import Badge from "./components/Badge";
+import Button from "./components/Button";
+import Filter from "./components/Filter";
+import StoryCard from "./components/StoryCard";
+
+const customTheme: Theme = extendTheme({
+  fonts: {
+    body: `"Comic Sans MS"`,
```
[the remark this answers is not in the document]
gaoxk: LOL ty for pointing this out, this is a remnant of sillier testing which chakra

frontend/src/components/homepage/Filter.tsx
ZhangCreations: Is there a reason we need to use the useStyleConfig hook here instead of just migrating the CSS to Chakra?

frontend/src/components/homepage/Filter.tsx
```diff
@@ -40,43 +51,49 @@ const Filter = ({
...
     );
   }
+  const filterStyle = useStyleConfig("Filter");
   return (
-    <div className="filter">
-      <h1>Filter Stories</h1>
-      <div>
-        <h4>TRANSLATION LANGUAGE</h4>
-        <select
+    <Flex sx={filterStyle}>
```
ZhangCreations, line 56: Not sure we should be using the `sx` prop like this, generally think it's bad practice in ChakraUI and whatever can be included in a theme should be.

frontend/src/components/homepage/StoryCard.tsx
```diff
@@ -137,9 +144,10 @@ const StoryCard = ({
     }
     return storyTranslationId ? assignReviewer : assignTranslator;
   };
+  const storyCardStyle = useStyleConfig("StoryCard");
 
   return (
-    <div id={`story-${storyId}`} className="story-card">
+    <Flex id={`story-${storyId}`} sx={storyCardStyle}>
```
ZhangCreations, line 150: Again, not sure about the use of `sx` here

frontend/src/components/homepage/StoryCard.tsx
```diff
@@ -148,24 +156,35 @@ const StoryCard = ({
...
-        <SecondaryButton onClick={previewBook} displayText="preview book" />
-      </div>
-    </div>
+        >
+          {primaryBtnText()}
+        </Button>
+        <Button
+          colorScheme="blue"
+          variant="outline"
+          size="wide"
```
ZhangCreations, line 181: If these traits will be common to a lot of buttons, we should be able to include this in the `customTheme`

frontend/src/theme/colors.ts
ZhangCreations: In the past, I've worked with a much simpler colour scheme with fewer colours so we could name things like "background" and "titles" etc.  If we coordinate with design on numbers for each tone i.e blue.100 the designers can note exactly which shade they're using in Figma which would be pretty fluid for devs.

ZhangCreations (changes requested):
Looks good to me! I just have a few small things to address.

frontend/src/theme/colors.ts
gaoxk: Good suggestion, sent a message in #planetread-design about it

frontend/src/components/homepage/Filter.tsx
gaoxk: I moved the css to Chakra (`themes/components/Filter`) for now and load it here through useStyleConfig 
I initially tried to put this css into a theme for `Flex` but `Flex` & `Box` don't accept variants and have to be loaded like this ([source](https://github.com/chakra-ui/chakra-ui/issues/2456))
(for context, the original component with `className="filter"` was a flex `div`, hence the desire to convert it to `Flex`)

gaoxk: The unfortunate `sx` usage is also a result of `useStyleConfig`

e-wai:
chakra review

General appearance comments. Not sure what belongs within this PR -> 
- the logo at the top left is way too small
- width issue still exists (scroll right/left)
- some spacing issues below text lines (inconsistent and/or too small)
- probably some changes to be made on the story translation content editing page
- still a lot of outstanding changes from usability for later?

frontend/src/components/homepage/StoryList.css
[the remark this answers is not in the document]
e-wai: Can this be changed to `backgroundSurfaceLofi`?

e-wai (changes requested):
I think it's mostly good to go, but could you clarify (whether through tickets or just a list) what the path forward is for the rest of the Chakra UI implementation + with the usability testing? Is it separate tickets for each page?

gaoxk:
> General appearance comments

~not to push the issue down the road, but also to totally push the issue down the road~ Will be addressed in hifi migration tickets

> Can this be changed to backgroundSurfaceLofi?

Unfortunately not since it's a plain css doc. Chakra wont work it's magic

e-wai (commented):
Looks good to me, but I'll leave the final word to @ZhangCreations since he has more context on Chakra.

ZhangCreations (approved):
Looks good to merge but just a comment that the pages don't seem to scale properly, not sure if this is something we're thinking about,.

PR #48, opened 2021-07-16
Story review platform/create comment

lynn-kim:
## Notion ticket link

[Create new / reply to comment](https://www.notion.so/uwblueprintexecs/Create-new-or-reply-to-comment-27fecc42fa6d4e05a91eeddee1e12245)

## Implementation description
* added a graphql mutation to create a new comment 
* created a comment request and response DTO

## Steps to test
1. make sure you have the comments table in your DB and have populated the `story_translation_contents` table (if not, then set `erase_db_and_sync = True`) 
2. go to http://localhost:5000/graphql and test mutation ([sample query](http://localhost:5000/graphql?variables=null&operationName=createComment&query=mutation%20createComment%7B%0A%20createComment(commentData%3A%7Btime%3A%20123456%2C%20userId%3A%201%2C%20content%3A%22hello%22%2C%20storyTranslationContentId%3A%202%2C%20commentIndex%3A0%2C%20resolved%3Afalse%7D)%20%7B%0A%20%20ok%0A%20%20comment%7B%0A%20%20%20%20content%0A%20%20%20%20time%0A%20%20%20%20userId%0A%20%20%20%20id%0A%20%20%20%20storyTranslationContentId%20%0A%20%20%7D%0A%7D%0A%7D ))
3. try with different values

## What should reviewers focus on?
* `backend/python/app/graphql/mutations/comment_mutation.py`
* `backend/python/app/graphql/types/comment_type.py`
* should comment time be stored as DateTime? 

## Checklist
- [ x ] My PR name is descriptive and in imperative tense
- [ x ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ x ] For backend changes, I have run the appropriate linters:  `docker exec -it <backend-container-id> /bin/bash -c "black . && isort --profile black ."`
- [ x ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #50, opened 2021-07-18
Fix translation content returned as null bug for story translations queries

ArmanyaDalmia:
## Notion ticket link

[Return translation content on storyTranslationsByUser and storyTranslationsAvailableForReview](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=6a0636cf217e423685acc3909a348d91)

## Implementation description
* Use two separate queries, one for the stories based on desired criteria and the other for the story translations with translation content
* Join the the two queries together in a dict and return a list of these dicts

## Steps to test
1. Run the `storyTranslationsByUser` and `storyTranslationsAvailableForReview` queries and observe that the `translationContents` field is not null

* Whether this is the best way to get the translation contents as it requires an additional query per row of the main one

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -97,18 +97,13 @@ class StoryService(IStoryService):
 
     def get_story_translations(self, user_id, translator, language, level):
         try:
-            return (
+            story_translations = (
                 db.session.query(
                     Story.id.label("story_id"),
                     Story.title.label("title"),
                     Story.description.label("description"),
                     Story.youtube_link.label("youtube_link"),
                     Story.level.label("level"),
-                    StoryTranslation.id.label("id"),
-                    StoryTranslation.language.label("language"),
-                    StoryTranslation.stage.label("stage"),
-                    StoryTranslation.translator_id.label("translator_id"),
-                    StoryTranslation.reviewer_id.label("reviewer_id"),
                 )
                 .join(StoryTranslation, Story.id == StoryTranslation.story_id)
                 .filter(
@@ -118,7 +113,16 @@ class StoryService(IStoryService):
                 )
                 .filter(StoryTranslation.language == language if language else True)
                 .filter(Story.level <= level if level else True)
+                .all()
             )
+            translationsList = []
+            for story_translation in story_translations:
+                story_translation_dict = story_translation._asdict()
+                translationsList.append({
+                    **StoryTranslation.query.get(story_translation_dict["story_id"]).to_dict(include_relationships=True),
+                    **story_translation._asdict(),
+                })
+            return translationsList
         except Exception as error:
             self.logger.error(str(error))
             raise error
@@ -222,18 +226,13 @@ class StoryService(IStoryService):
 
     def get_story_translations_available_for_review(self, language, level):
         try:
-            return (
+            story_translations = (
                 db.session.query(
                     Story.id.label("story_id"),
                     Story.title.label("title"),
                     Story.description.label("description"),
                     Story.youtube_link.label("youtube_link"),
                     Story.level.label("level"),
-                    StoryTranslation.id.label("id"),
-                    StoryTranslation.language.label("language"),
-                    StoryTranslation.stage.label("stage"),
-                    StoryTranslation.translator_id.label("translator_id"),
-                    StoryTranslation.reviewer_id.label("reviewer_id"),
                 )
                 .join(StoryTranslation, Story.id == StoryTranslation.story_id)
                 .filter(Story.level <= level)
@@ -241,6 +240,14 @@ class StoryService(IStoryService):
                 .filter(StoryTranslation.reviewer_id == None)
                 .all()
             )
+            translationsList = []
+            for story_translation in story_translations:
+                story_translation_dict = story_translation._asdict()
+                translationsList.append({
+                    **StoryTranslation.query.get(story_translation_dict["story_id"]).to_dict(include_relationships=True),
+                    **story_translation._asdict(),
+                })
+            return translationsList
 
         except Exception as error:
             self.logger.error(str(error))
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -97,15 +97,8 @@ class StoryService(IStoryService):
 
     def get_story_translations(self, user_id, translator, language, level):
         try:
-            story_translations = (
-                db.session.query(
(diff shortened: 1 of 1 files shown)


backend/python/app/services/implementations/story_service.py
```diff
@@ -225,25 +233,32 @@ def update_story_translation_contents(self, story_translation_contents):
...
                 .filter(StoryTranslation.language == language)
                 .filter(StoryTranslation.reviewer_id == None)
                 .all()
             )
+            translationsList = []
+            for story_translation in story_translations:
+                story_translation_dict = story_translation._asdict()
+                translationsList.append(
+                    {
+                        **StoryTranslation.query.get(
```
gaoxk, line 255: Is it possible to use only one query for this entire function? (and `get_story_translations` and `get_story_translation`. `get_story_translation` works but should really only be 1 query.) This might be inspiring/helpful: https://docs.sqlalchemy.org/en/14/tutorial/orm_related_objects.html#using-relationships-in-queries

You might be able to do some fancy stuff with `Story.join(StoryTranslation).join(StoryTranslationContents)`

gaoxk (changes requested):

backend/python/app/services/implementations/story_service.py
e-wai: ^ in line with this, can we look into whether there's a `.all` that exists? Seems more efficient than iterating

backend/python/app/graphql/mutations/auth_mutation.py
```diff
@@ -161,7 +161,7 @@ def mutate(root, info, first_name, last_name, email, password):
                     last_name=last_name,
                     email=email,
                     role="User",
-                    password=password
+                    password=password,
```
e-wai, line 164: Heads-up that https://github.com/uwblueprint/planet-read/pull/48 makes similar linting changes to you. (I think they're all the same and so merging should then go smoothly? ? ? don't remember)

gaoxk: should require pretty trivial rebase on either #48 or #50

backend/python/app/services/implementations/story_service.py
```diff
@@ -222,26 +231,29 @@ def update_story_translation_contents(self, story_translation_contents):
...
+                StoryTranslation.query.join(Story, Story.id == StoryTranslation.story_id)
+                .filter(Story.level <= level)
+                .filter(StoryTranslation.language == language)
+                .filter(StoryTranslation.reviewer_id == None)
+                .order_by(Story.id)
+                .all()
+            )
+
+            for i in range(len(stories)):
+                stories[i] = {**stories[i].to_dict(), **story_translations[i].to_dict(include_relationships=True)}
```
gaoxk, line 253: drive by comment: something like this might fit your situation
https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-taking-union-of-dictiona

backend/python/app/services/implementations/story_service.py
```diff
@@ -97,36 +97,43 @@ def create_translation(self, entity):
...
+            return stories
+            
         except Exception as error:
             self.logger.error(str(error))
             raise error
 
     def get_story_translation(self, id):
         try:
-            # TODO: Condense into singular query
-            translation = StoryTranslation.query.get(id)
```
gaoxk, line 137: nit: remove empty line

backend/python/app/services/implementations/story_service.py
```diff
@@ -135,20 +142,35 @@ def get_story_translation(self, id):
...
                 .filter(StoryTranslation.id == id)
-                .one()
-                ._asdict()
+                .all()
             )
 
-            story_details["num_translated_lines"] = self._get_num_translated_lines(
-                translation.translation_contents
+            response = {**story_details[0]._asdict()}
+            response.pop("content_id")
```
gaoxk, line 161: Rather than pop-ing these attributes, what about removing lines 150-152?

gaoxk (changes requested):
Nice! Looks good so far

backend/python/app/services/implementations/story_service.py
```diff
@@ -97,36 +97,43 @@ def create_translation(self, entity):
...
-                    Story.title.label("title"),
-                    Story.description.label("description"),
-                    Story.youtube_link.label("youtube_link"),
-                    Story.level.label("level"),
-                    StoryTranslation.id.label("id"),
-                    StoryTranslation.language.label("language"),
-                    StoryTranslation.stage.label("stage"),
-                    StoryTranslation.translator_id.label("translator_id"),
-                    StoryTranslation.reviewer_id.label("reviewer_id"),
+            stories = (
```
gaoxk, line 100: I like what you did for `get_story_translation` to use just 1 query - what happens if you use a similar approach here?

ArmanyaDalmia: It should be doable but would require a lot of post-processing since it would entail combining all the storytranslationcontent rows with the same `story_translation_id` together into a dict and doing this for all rows

backend/python/app/services/implementations/story_service.py
ArmanyaDalmia: Lines 150-152 are required as each result from `StoryTranslationContent` in the `translationContent` part of the gql query needs those attributes. Popping them is just to create the initial dict that needs to be returned, with the for loop creating the list of dicts for the `translationContent` field

gaoxk: Interesting, to clarify, what is at `response["content_id"]` and `response["line_index"]` and `response["translation_content"]` at this point of the code? Is it just empty? looking at this function again, I'm surprised these fields exist on the response

backend/python/app/services/implementations/story_service.py
gaoxk: understood, happy with this state then 👍

backend/python/app/services/implementations/story_service.py
ArmanyaDalmia: It would just be the values for the first result of the sql query (ie. the details for the first translated line of specified story_translation) since the returned response needs all the other key-value pairs and then needs lines 150-152 as a list of dicts for the `translationContent` field

gaoxk (approved):
LGTM! 👍

PR #51, opened 2021-07-19
Refactor saving mechanism into Autosave component

e-wai:
## Notion ticket link

[Use react hooks to avoid per-keystroke auto save](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=33d890f91c9d4875b59c8a627bf320ff)

## Implementation description
* Based on https://www.synthace.com/autosave-with-react-hooks/
* Create debounce function in utils, create Autosave function for the updateStoryTranslationContents call. Note: replaced updateStoryTranslationContentById, as theoretically multiple lines could've been changed within a second. 
* Introduce `changedStoryLines` for state -> to store the changed lines that have to be mutated. Using a Map so that within a second, only 1 call per line changed

## Steps to test
1. Make edits on translation page. Refresh/check database to make sure saves are being saved

autosave fix

* Is this a reasonable timeout delay?
* Probably will need some error handling? -> some success / failure indicator

## Checklist

### DIFF
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -51,6 +51,9 @@ const TranslationPage = () => {
   const [translatedStoryLines, setTranslatedStoryLines] = useState<StoryLine[]>(
     [],
   );
+  const [changedStoryLines, setChangedStoryLines] = useState<
+    Map<number, StoryLine>
+  >(new Map());
   const [percentageComplete] = useState(25);
 
   const arrayIndex = (lineIndex: number): number =>
@@ -75,9 +78,17 @@ const TranslationPage = () => {
     lineIndex: number,
   ) => {
     const updatedContentArray = [...translatedStoryLines];
-    updatedContentArray[arrayIndex(lineIndex)].translatedContent = newContent;
+    const index = arrayIndex(lineIndex);
+    updatedContentArray[index].translatedContent = newContent;
 
     setTranslatedStoryLines(updatedContentArray);
+    setChangedStoryLines(
+      changedStoryLines.set(lineIndex, updatedContentArray[index]),
+    );
+  };
+
+  const clearUnsavedChangesMap = () => {
+    setChangedStoryLines(new Map());
   };
 
   useQuery(GET_STORY_CONTENTS(storyId, storyTranslationId), {
@@ -139,7 +150,10 @@ const TranslationPage = () => {
           </div>
         </div>
       </div>
-      <Autosave storylines={translatedStoryLines} />
+      <Autosave
+        storylines={Array.from(changedStoryLines.values())}
+        onSuccess={clearUnsavedChangesMap}
+      />
     </div>
   );
 };
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { gql, useMutation, useQuery } from "@apollo/client";
+import { gql, useQuery } from "@apollo/client";
 
 import "./TranslationPage.css";
 import { useParams } from "react-router-dom";
@@ -8,17 +8,11 @@ import EditableCell from "../translation/EditableCell";
 import TranslationProgressBar from "../translation/TranslationProgressBar";
 import CheckmarkIcon from "../../assets/checkmark.svg";
 import CommentIcon from "../../assets/comment_no_number.svg";
+import Autosave, { StoryLine } from "../translation/Autosave";
 
 type TranslationPageProps = {
   storyIdParam: string | undefined;
   storyTranslationIdParam: string | undefined;
-};
-
-type StoryLine = {
-  lineIndex: number;
-  originalContent: string;
-  translatedContent?: string;
-  storyTranslationContentId?: number;
 };
 
 type Content = {
@@ -45,27 +39,6 @@ const GET_STORY_CONTENTS = (storyId: number, storyTranslationId: number) => gql`
   }
 `;
 
-type UpdateTranslationResponse = {
-  story: { lineIndex: number };
-};
-const UPDATE_TRANSLATION = gql`
-  mutation updateStoryTranslationContentById(
-    $storyTranslationContentId: Int!
-    $newTranslationContent: String!
-  ) {
-    updateStoryTranslationContentById(
-      storyTranslationContentData: {
-        id: $storyTranslationContentId
-        translationContent: $newTranslationContent
-      }
-    ) {
-      story {
-        lineIndex
-      }
-    }
-  }
-`;
-
 const TranslationPage = () => {
   const {
     storyIdParam,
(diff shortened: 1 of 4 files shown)


frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -101,35 +73,22 @@ const TranslationPage = () => {
...
 
-        setTranslatedStoryLines(updatedContentArray);
-      } else {
-        handleError("Unable to save translation.");
-      }
-    } catch (err) {
-      handleError(err ?? "Error occurred, please try again.");
-    }
+  const clearUnsavedChangesMap = () => {
+    setChangedStoryLines(new Map());
```
e-wai, line 91: There may be an unnecessary render triggered by this?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -78,12 +51,11 @@ const TranslationPage = () => {
   const [translatedStoryLines, setTranslatedStoryLines] = useState<StoryLine[]>(
     [],
   );
+  const [changedStoryLines, setChangedStoryLines] = useState<
```
gaoxk, line 54: @jennchenn FYI - Emma is recording what lines are changed. This has overlap with your current ticket

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -190,6 +150,10 @@ const TranslationPage = () => {
           </div>
         </div>
       </div>
+      <Autosave
```
gaoxk, line 153: out of curiosity, how do you think we can scale autosaving if user testing finds that comments should be auto save on edit?

frontend/src/components/pages/TranslationPage.tsx
gaoxk: Hmmm I think this is inevitable. I think the only way to reduce rerenders is to make `changedStoryLines` non state, but in the current code on this pr, `changedStoryLines` needs to be in react state for the `Autosave` to debounce properly.

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -190,6 +150,10 @@ const TranslationPage = () => {
...
+        storylines={Array.from(changedStoryLines.values())}
```
gaoxk, line 154: It's a shame that we have to set up an additional state `changedStoryLines` when it _feels_ like we could figure out the state changes and debounce with just `TranslatedStoryLines`. But I can't think of a different way to do this :// It's not a big deal since @jennchenn is likely going to need to do something similar like this anyways, but let me know if you have any thoughts.

frontend/src/utils/DebounceUtils.ts
```diff
@@ -0,0 +1,19 @@
+// Inspiration from  https://levelup.gitconnected.com/debounce-in-javascript-improve-your-applications-performance-5b01855e086
```
gaoxk, line 1: great resource!

frontend/src/components/translation/Autosave.tsx
```diff
@@ -0,0 +1,75 @@
...
+      });
+
+      try {
+        const result = await updateTranslation({
+          variables: { storyTranslationContents },
+        });
+
+        if (result.data == null) {
+          handleError("Unable to save translation");
+        } else if (linesToUpdate.length > 0) {
```
gaoxk, line 56: This condition strikes me as a little odd - I believe `linesToUpdate.length > 0` will always be true if `linesToUpdate `was passed with any length - did you mean something like `result.data.length > 0`

gaoxk (commented):
Love to see the demo with the network panel open! Left some questions

frontend/src/components/pages/TranslationPage.tsx
e-wai: As in per keystroke? Wouldn't we just revert back to the original state?

frontend/src/components/pages/TranslationPage.tsx
e-wai: Yep, I couldn't think of a way because the UI is being updated at a different rate than the backend now so TranslatedStoryLines needs to be updated immediately. 

I was trying to imagine a world where the Autosave component could "cache" the story lines to "diff" what needs to be updated in each request, but even if that worked, it would need to iterate through the entire array to find out what changed, I think???

frontend/src/components/translation/Autosave.tsx
e-wai: Hmm don't remember why I added this, might've been a "make it work" type condition because at one point requests were firing more than expected. Will check.

frontend/src/components/pages/TranslationPage.tsx
gaoxk: Ideally we would avoid per keystroke saving again. Would we be able to use another `<Autosave>` for comment changes?

e-wai: Ohh for comments. I'm thinking the easiest way would be that `changedStoryLines` becomes something like:
```json
changesToSave = {
  "storyLines": StoryLine [],
  "comments": Comment [],
}
```
and is still passed as a prop. Call executes none, one, or both mutations as necessary? Alternatively, we could introduce another state variable exclusively for comments and pass both into the Autosave component as props. The latter would probably be better for legibility, but I'm not sure whether it's worth introducing another state variable when changesToSave does make sense as a group + we want to execute mutations together in calls.

frontend/src/components/translation/Autosave.tsx
e-wai: I think it's just to prevent unnecessary calls that would prompt a re-render on parent page. Don't think result.data.length is anything useful?

Changed it up so that there is a length check at the very beginning. Results with no changes if so.

e-wai:
autosave revision

frontend/src/components/pages/TranslationPage.tsx
gaoxk: This sounds good to me - the only downside in UX is if the user edits text and then immediately edits a comment, the text save would be delayed until the comment changes stop long enough to be saved.
We currently have two different mutations but I think that's fine - the case of both story translation content and comments being edited within a 1000ms period is probably pretty rare

frontend/src/components/pages/TranslationPage.tsx
gaoxk: "cache"ing would be interesting - I'm not sure if saving or diff-ing old state is an ugly react pattern. AutoSave's "cache" state would be reliant on the previous props it received, but accessing "previous props" certainly isn't a react-like pattern since the entire component has been rerendered

frontend/src/components/translation/Autosave.tsx
gaoxk: Interesting, thanks for humoring me
Happy with the current code

ZhangCreations (commented):
Looks good to me!

gaoxk (approved):

frontend/src/components/pages/TranslationPage.tsx
e-wai: Agree it's an edge case still worth looking at. The mutations should be together on the frontend I think? but will need distinct resolving + database locks in backend

```
mutation {
  mutation1() {}
  mutation2() {}
}
```

PR #52, opened 2021-07-21
Integrate Translation Process Indicator

jennchenn:
## Notion ticket link
[Integrate Translation Process Indicator](https://www.notion.so/uwblueprintexecs/integrate-translation-process-indicator-9b9e877c7dce428ab45a9b70a33e0621)

## Implementation description
* Set progress bar value on first render based on return value from `storyTranslationById` query
* On user updates, determine if user is adding/removing translation lines and update progress bar accordingly 
* Branch based off of #51 

progress bar

## Steps to test
1. Seed database with stories
2. Navigate to `http://localhost:3000/stories`; select a story and click "translate book"
3. Verify that progress bar originally is at 0%
4. Enter translation: verify that progress bar increments accordingly
5. Delete translation: verify that progress bar decrements accordingly
6. Refresh the page: verify that progress bar value is accurately reflected

* Ensure whitespaces aren't counted as translated lines
* Verify that percentage is within 0%-100% at all times

## Checklist

### DIFF
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -34,7 +34,8 @@ const GET_STORY_CONTENTS = (storyId: number, storyTranslationId: number) => gql`
         id
         lineIndex
         content: translationContent
-      }
+      },
+      numTranslatedLines
     }
   }
 `;
@@ -54,7 +55,9 @@ const TranslationPage = () => {
   const [changedStoryLines, setChangedStoryLines] = useState<
     Map<number, StoryLine>
   >(new Map());
-  const [percentageComplete] = useState(25);
+  const [numTranslatedLines, setNumTranslatedLines] = useState(0);
+  const [numStoryLines, setNumStoryLines] = useState(0);
+  const [percentageComplete, setPercentageComplete] = useState(0);
 
   const arrayIndex = (lineIndex: number): number =>
     lineIndex - translatedStoryLines[0].lineIndex;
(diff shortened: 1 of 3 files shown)


frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -54,7 +55,9 @@ const TranslationPage = () => {
   const [changedStoryLines, setChangedStoryLines] = useState<
     Map<number, StoryLine>
   >(new Map());
-  const [percentageComplete] = useState(25);
+  const [numTranslatedLines, setNumTranslatedLines] = useState(0);
+  const [numStoryLines, setNumStoryLines] = useState(0);
```
e-wai, line 59: Is `numStoryLines` necessary? Is it ever different from the length of `translatedStoryLines`, would it ever need to trigger state change?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -73,14 +76,35 @@ const TranslationPage = () => {
...
     lineIndex: number,
   ) => {
     const updatedContentArray = [...translatedStoryLines];
     const index = arrayIndex(lineIndex);
-    updatedContentArray[index].translatedContent = newContent;
 
+    if (
+      // user deleted translation line
+      !newContent.trim() &&
+      translatedStoryLines[index]?.translatedContent?.trim()
```
e-wai, line 97: Hmm why a null check on `translatedStoryLines` and `translatedContent`? The only time I would expect .translatedContent to be null is on the super initial render, but it should already exist here

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -54,7 +55,9 @@ const TranslationPage = () => {
   const [changedStoryLines, setChangedStoryLines] = useState<
     Map<number, StoryLine>
   >(new Map());
-  const [percentageComplete] = useState(25);
+  const [numTranslatedLines, setNumTranslatedLines] = useState(0);
+  const [numStoryLines, setNumStoryLines] = useState(0);
+  const [percentageComplete, setPercentageComplete] = useState(0);
```
e-wai, line 60: Does percentageComplete need to be a state variable? It won't trigger any changes not already found by the `numTranslatedLines`, and can be calculated as a const or even in line, I think

e-wai (changes requested):
Tiny bit weird to me to have num translated lines be owned both by the backend and the frontend, but makes sense for our use cases. Couple of questions 👍

frontend/src/components/pages/TranslationPage.tsx
jennchenn: nope will remove that!

frontend/src/components/pages/TranslationPage.tsx
jennchenn: I removed the null check on `translatedStoryLines` but get an "Object is possibly undefined" error without the optional chaining for `translatedContent`

e-wai: Yeah I'd argue that it should be `translatedContent!!` since something is very wrong if it's null? Don't know if it's bad practice but I'd rather it avoid "absorbing" errors

e-wai (approved):
Responded to the comment again, but doesn't block I think

e-wai (changes requested):
Oh jokes, one more thing - should the right bar be sticky?

gaoxk (approved):

e-wai (approved):

gaoxk (approved):

PR #37, opened 2021-06-24
Add filter to homepage

gaoxk:
## Notion ticket link

[add filter](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=4a47edc475b34bb99b25ff1fae2a5507&p=ca91e0de049d45fc9daf0ca5953e410e)

ezgif com-gif-maker

## Implementation description
* added approved_languages to user context
* added filter component
* hooked up state + user context in Homepage with Filter

note: 
- renamed `story` folder to `homepage`
- **the home page will break for users with no approved languages set**

## Steps to test
1. Insert a user with approved languages `'{"HINDI":3, "FRENCH":2, "BENGALI":4}'`. make a second user, but log in as the first user.
2.  insert some NGO friendly stories
```
INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('Goodnight, Tinku!', 'Tinku, a little pup at Mangu\'s farm is not sleepy at all. He decides to step out into the night and meets many interesting animals.', 'https://www.youtube.com/watch?v=evHY-2oSokY', 2, '{"HINDI", "FRENCH"}');
INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('The Flying Elephant', 'Find out how a flying elephant changes the life of a poor farmer.', 'https://www.youtube.com/watch?v=hh_4Axs3ZVQ', 4, '{"BENGALI"}');
INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('The Lion and the Fox', 'The smart lion thinks he can use the fox to bring him food every day. But the fox is smarter.', 'https://www.youtube.com/watch?v=pFDXCXVvtRo', 3, '{"BENGALI"}');
INSERT INTO stories (title, description, youtube_link, level, translated_languages) VALUES ('A Book for Puchku', 'Puchku has run out of books to read. Then she discovers more books in the top shelf of the library bookcase. But Puchku is small, and the bookcase tall. How will she ever get to her beloved books?', 'https://www.youtube.com/watch?v=PKxMjO4HPvE', 3, '{}');
```
3. Insert the translations
```
INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (1, 'HINDI', 'TRANSLATE', 1);
INSERT INTO story_translations (story_id, language, stage, translator_id) VALUES (1, 'FRENCH', 'TRANSLATE', 1);
[...]

PR #55, opened 2021-07-29
Implemented Undo/Redo Functionality

ZhangCreations:
## Notion ticket link
[Allow translation undo and redo](https://www.notion.so/uwblueprintexecs/Allow-translation-undo-and-redo-7f36573f90a040748c6cb81f870a7dd1)

## Implementation description
* When a user enters a character, takes a snapshot of the current state and pushes it onto a stack.
* Undo -> Pop from undo stack and add current state to redo
* Redo -> Pop from  redo stack and add current state to undo
* Not debounced since figuring out exactly when to snapshot can be tricky esp in conjunction with autosave, this is more MVP for undo/redo.

## Steps to test
1. Launch the app and go to the translation page for some story
2. Type some stuff into the translation area, perhaps multiple text areas
3. Press the undo button (not using ctrl + z, that's built into textarea so that would always work), this should undo the last press you've made
4. Try redoing
6. Send a photo of the most derelict/broken bike you can find to @andrewyguo on slack to make him suffer. 

https://user-images.githubusercontent.com/53587708/127574770-9105506b-9888-4b92-abb6-08223b4523dd.mp4

* Note: The buttons are not styled yet since I think they would need to be re-implemented once [Add Chakra UI #49](https://github.com/uwblueprint/planet-read/pull/49) is merged in.
* Make sure the edge cases work (i.e when you undo/redo all the work you've done in a session) I've had some weird finicky stuff with it and I'm pretty sure it no longer breaks but just saying that I wouldn't be the most surprised.
* When the user starts typing after a ton of undos, they wipe all the redo screenshots, does this make sense?

## Checklist

<p align="center">
  
</p>

### DIFF
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -20,6 +20,12 @@ type Content = {
   lineIndex: number;
   content: string;
 };
+
+interface HistoryStack {
+  Undo: Array<{ lineIndex: number; content: string }>;
+  Redo: Array<{ lineIndex: number; content: string }>;
+}
+
 const GET_STORY_CONTENTS = (storyId: number, storyTranslationId: number) => gql`
   query {
     storyById(id: ${storyId}) {
@@ -56,8 +62,20 @@ const TranslationPage = () => {
   >(new Map());
   const [numTranslatedLines, setNumTranslatedLines] = useState(0);
 
+  const [
+    versionHistoryStack,
+    updateVersionHistoryStack,
+  ] = useState<HistoryStack>({ Undo: [], Redo: [] });
+
   const arrayIndex = (lineIndex: number): number =>
     lineIndex - translatedStoryLines[0].lineIndex;
+
+  const deepCopy = (lines: Object) => {
+    // This is a funky method to make deep copies on objects with primative values
+    // https://javascript.plainenglish.io/how-to-deep-copy-objects-and-arrays-in-javascript-7c911359b089
+    // Should probably go under some util
+    return JSON.parse(JSON.stringify(lines));
+  };
 
   // TODO replace with real logic
   const translationStatusIcon = (): JSX.Element | null => {
@@ -73,13 +91,19 @@ const TranslationPage = () => {
     }
   };
 
+  const updateVersionHistory = (content: string, lineIndex: number) => {
+    updateVersionHistoryStack({
+      Undo: [...deepCopy(versionHistoryStack.Undo), { lineIndex, content }],
+      Redo: [],
+    });
+  };
+
   const onChangeTranslationContent = async (
     newContent: string,
     lineIndex: number,
   ) => {
     const updatedContentArray = [...translatedStoryLines];
     const index = arrayIndex(lineIndex);
-
     if (
       // user deleted translation line
       !newContent.trim() &&
@@ -98,6 +122,65 @@ const TranslationPage = () => {
     setChangedStoryLines(
       changedStoryLines.set(lineIndex, updatedContentArray[index]),
     );
+  };
+
+  const onUserInput = async (newContent: string, lineIndex: number) => {
+    const oldContent =
+      translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+    if (oldContent !== undefined) {
+      updateVersionHistory(oldContent, lineIndex);
+    }
+    onChangeTranslationContent(newContent, lineIndex);
+  };
+
+  console.log(versionHistoryStack);
+
+  const undoChange = () => {
+    if (versionHistoryStack.Undo.length > 0) {
+      const { lineIndex, content: newContent } = versionHistoryStack.Undo[
+        versionHistoryStack.Undo.length - 1
+      ];
+      const oldContent =
+        translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+      if (oldContent !== newContent) {
+        const newHistory = {
+          Undo: deepCopy(versionHistoryStack.Undo.slice(0, -1)),
+          Redo: [
+            ...deepCopy(versionHistoryStack.Redo),
+            {
+              lineIndex,
+              content: oldContent,
+            },
+          ],
+        };
+        updateVersionHistoryStack(newHistory);
+        onChangeTranslationContent(newContent, lineIndex);
+      }
+    }
+  };
+
+  const redoChange = () => {
+    if (versionHistoryStack.Redo.length > 0) {
+      const { lineIndex, content: newContent } = versionHistoryStack.Redo[
+        versionHistoryStack.Redo.length - 1
+      ];
+      const oldContent =
+        translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+      if (oldContent !== newContent) {
+        const newHistory = {
+          Undo: [
+            ...deepCopy(versionHistoryStack.Undo),
+            {
+              lineIndex,
+              content: oldContent,
+            },
+          ],
+          Redo: deepCopy(versionHistoryStack.Redo.slice(0, -1)),
+        };
+        updateVersionHistoryStack(newHistory);
+        onChangeTranslationContent(newContent, lineIndex);
+      }
+    }
   };
 
   const clearUnsavedChangesMap = () => {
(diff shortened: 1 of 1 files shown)


andrewyguo:
1/10 Meme, not approved

e-wai (commented):
(Haven't taken a look at everything - just a note that I tested the undo / redo between different sites and VScode, and behaviour is not consistent, however, most don't go character by character)

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -100,6 +124,63 @@ const TranslationPage = () => {
     );
   };
 
+  const onUserInput = async (newContent: string, lineIndex: number) => {
+    const oldContent =
+      translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+    if (oldContent !== undefined) {
```
gaoxk, line 130: I believe you can skip this check if 129 has a [non null assertion operator](https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci)
something like
```
const oldContent =
      translatedStoryLines[arrayIndex(lineIndex)].translatedContent!;
```

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -20,6 +20,12 @@ type Content = {
   lineIndex: number;
   content: string;
 };
+
+interface HistoryStack {
+  Undo: Array<{ lineIndex: number; content: string }>;
+  Redo: Array<{ lineIndex: number; content: string }>;
+}
```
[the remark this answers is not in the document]
gaoxk: the conversation we had was about props (#37). Generally speaking, the main difference is that interfaces `An interface can have multiple merged declarations, but a type alias for an object type literal cannot.` i.e. typescript will merge redeclared interfaces
```
interface Box {
  height: number;
  width: number;
}
interface Box {
  scale: number;
}
let box: Box = { height: 5, width: 6, scale: 10 };
```
[interfaces vs types](https://stackoverflow.com/questions/37233735/typescript-interfaces-vs-types)
[merged declarations](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)

So, take away:
nit: convert this to type. Leaving the door for accidental merging is a code smell

gaoxk (changes requested):
Nice! Didn't find any edgecases.

> When the user starts typing after a ton of undos, they wipe all the redo screenshots, does this make sense?

yep!

> Not debounced since figuring out exactly when to snapshot can be tricky esp in conjunction with autosave

Good to know - let's make a follow up pr

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -56,9 +62,21 @@ const TranslationPage = () => {
   >(new Map());
   const [numTranslatedLines, setNumTranslatedLines] = useState(0);
 
+  const [
+    versionHistoryStack,
+    updateVersionHistoryStack,
```
e-wai, line 67: Out of curiosity - is there a "good practice" around calling this `setVersionHistoryStack`? or doesn't really matter

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -100,6 +124,63 @@ const TranslationPage = () => {
...
+  };
+
+  const redoChange = () => {
+    if (versionHistoryStack.Redo.length > 0) {
+      const { lineIndex, content: newContent } = versionHistoryStack.Redo[
+        versionHistoryStack.Redo.length - 1
+      ];
+      const oldContent =
+        translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+      if (oldContent !== newContent) {
```
e-wai, line 167: In what case would we get `oldContent === newContent`?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -73,13 +91,19 @@ const TranslationPage = () => {
     }
   };
 
+  const updateVersionHistory = (content: string, lineIndex: number) => {
```
e-wai, line 94: This is very open for discussion, but this abstraction feel strange to me since this "level" of change is kept within the function body in the redo and undo methods and also, the name implies this method would work for the redo/undo scenarios.

Basically, I don't think giving its own method is super helpful. Thoughts?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -56,9 +62,21 @@ const TranslationPage = () => {
   >(new Map());
   const [numTranslatedLines, setNumTranslatedLines] = useState(0);
 
+  const [
+    versionHistoryStack,
+    updateVersionHistoryStack,
+  ] = useState<HistoryStack>({ Undo: [], Redo: [] });
+
   const arrayIndex = (lineIndex: number): number =>
```
e-wai, line 70: Not on you, I initially introduced this because I thought pagination would eventually be implemented. It sounds like we're going for infinite scroll, so `lineIndex === translatedStoryLines[0].lineIndex` always. (TL;DR, I abstracted too early, and I think it's worth removing this ASAP bc I'd rather not leave this stale piece to be used and context to be lost)

double-checking @gaoxk, pagination is a no?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -100,6 +124,63 @@ const TranslationPage = () => {
...
+  const undoChange = () => {
+    if (versionHistoryStack.Undo.length > 0) {
+      const { lineIndex, content: newContent } = versionHistoryStack.Undo[
+        versionHistoryStack.Undo.length - 1
+      ];
+      const oldContent =
+        translatedStoryLines[arrayIndex(lineIndex)].translatedContent;
+      if (oldContent !== newContent) {
+        const newHistory = {
+          Undo: deepCopy(versionHistoryStack.Undo.slice(0, -1)),
```
e-wai, line 145: Might be a slow night for me; could you explain why we need the `deepCopy`?

e-wai (dismissed):
Local test looks good to me 👍

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: Yeah, I named this thinking more in terms of how it only add/removes one thing at a time but it should probably be named `setVersionHistoryStack` based on what it actually does

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: I think a while back when writing this line I found that the state would double-save when the user first opened the page.

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: Yeah no worries, this took me forever to figure out. Basically without deepCopy, Typescript thinks its smart and stores a reference to the object rather than the actual contents, but that means that every object in the stack will update every time a new element is added.

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: Understandable and I totally see it now, I'll move this functionality into `onUserInput`.

gaoxk (changes requested):
Happy to merge after some quick nits!

P.s.  I like using `git commit --fixup` for addressing pr comments - it directly references the commit being fixed, and also auto squashes in other git circumstances (we currently always rebase and squash, but it's possible to merge into master instead.)
For this recent commit, it couldve been generated like `git commit --fixup 92d5ad2`

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -138,20 +137,27 @@ const TranslationPage = () => {
       const { lineIndex, content: newContent } = versionHistoryStack.Undo[
         versionHistoryStack.Undo.length - 1
       ];
+      console.log(
```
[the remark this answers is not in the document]
ZhangCreations: I totally saw your comment about `git commit --fixup` and thought "that's super cool, I'm gonna do that!" Then forgot to do it, but thanks for the suggestion I had no idea that existed.

gaoxk (approved):
🚢 excited to see it!

frontend/src/components/pages/TranslationPage.tsx
gaoxk: Woops thought I already replied, pagination is a no

PR #53, opened 2021-07-26
Add pytest

e-wai:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)
- tests for all services
- utilities?
- models?
- graphQL endpoints
- double check it's using the test_db -> will need to setup data for this
- flakiness of test_add_story()

## Implementation description
* 

## Steps to test
1.

## What should reviewers focus on?
* 

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] For backend changes, I have run the appropriate linters:  `docker exec -it <backend-container-id> /bin/bash -c "black . && isort --profile black ."`
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #63, opened 2021-08-11
Add pytest to backend

e-wai:
(This is the entire description of part 1 and 2.)

## Notion ticket link

[Add pytest](https://www.notion.so/uwblueprintexecs/Add-pytest-78637b82f2d9450eab9ef72dcf011290)
- This is part 1 of 2. Part 2 is available here: https://github.com/uwblueprint/planet-read/pull/69

## Usage
Test files are found by the following:
- Tests are in a `tests/` directory
- The test file is named `test_<something>.py`
- Each test is defined in a function whose prefix is `test_`:
```python
def test_<describe_the_test>():
  pass
```
The available fixtures currently are `app`, `db`, `client`, and `services`. If you pass the name of the fixture as a parameter of the test, it will execute fixture code before running your test and return or yield to you.
```python
def test_<describe_the_test>(app):
  result = app.run_something(True)
  assert(result)
```

## Implementation description
* Add testing configuration, which allows for better error log messages while testing, which is determined by the `config_name` passed to `create_app`. Normally, `create_app` is called from `server.py` which is the command that initiates the backend, but for testing it will be explicitly called if not created through the `app` fixture with `config_name = 'testing'`. This mostly just changes the db to be used.
* `conftest.py` has the fixtures we're using right now.
* Trivia: `app.app_context().push` is called same as in the development flow in the `models/__init__.py` which is called from the `create_app` flow and should therefore be available everywhere
* Trivia no. 2: ran into a ton of problems with parent packages not being found (run https://github.com/uwblueprint/planet-read/pull/53/ to observe the 😭) and restructuring made it clear where I could insert `__init__.py` files and eventually worked
* Added a `utils.sh` script because I'm lazy and like things abbreviated

## Steps to test
1. Rebuild (`docker-compose down -v`, `docker-compose up --build`) and recreate db 
2. Run the tests using `./utils.sh pytest`
3. Try out the other pytest command options: for a specific file, and for a file + name pattern match. Run it manually with `docker exec -it <backend-container-id> /bin/bash <pytest command>`
4. Verify that the tests db and general db refer to different instances: Make a change (whether via backend or frontend) to one of the tables, run the tests (which will drop and recreate everything), and check that your change has persisted. 

* Test structure: I think there are two schools of thoughts here, (1) that tests should live in a distinct `app/tests` directory that mirrors the file structure of the rest of the app, or (2) that tests should live right next to the actual code. I opted for (1) because it made it easy to add `__init__.py` files that made it work and I don't think imports look too, too bad. 
* Fixture placement: all our fixtures are in a single location right now, `conftest.py`. I think we should split this up as we go. Similar thought for helpers - we'll need somewhere to put them that isn't just one long file
* The one mutation test that was implemented, `test_story_mutation.py` does a lot of verifying, checking both the expected response type and that the expected changes were made to the db for each field. This causes some problems: there are some weird type mismatches and one is snake case vs. other in camel case, which with nesting, is a pain to sort out. Is this the right approach/extensiveness at which we want to test?
* Somewhat related, Graphene offers an option for snapshot testing graphene (see [docs](https://docs.graphene-python.org/en/latest/testing/#snapshot-testing)). I'm not sold on using it, but open to feedback -- it is much more concise, but somewhat hides our result from us.
* Read through the tests cases. This PR isn't focused on extensive test coverage (so naming isn't consistent, probably missing cases, and I definitely use error vs. exception interchangeably) and will grow as we implement more, but let me know if it all makes sense.

## Checklist

### DIFF
--- a/backend/python/requirements.txt
+++ b/backend/python/requirements.txt
@@ -37,7 +37,6 @@ Jinja2==2.11.2
 Mako==1.1.4
 MarkupSafe==1.1.1
 mysqlclient==2.0.3
-mongoengine==0.23.0
 msgpack==1.0.2
 mypy-extensions==0.4.3
 packaging==20.9
@@ -48,7 +47,6 @@ psycopg2==2.8.6
 pyasn1==0.4.8
 pyasn1-modules==0.2.8
 pycparser==2.20
-pymongo==3.11.3
 pyparsing==2.4.7
 pytest==6.2.4
 python-dateutil==2.8.1
--- a/backend/python/requirements.txt
+++ b/backend/python/requirements.txt
@@ -50,6 +50,7 @@ pyasn1-modules==0.2.8
 pycparser==2.20
 pymongo==3.11.3
 pyparsing==2.4.7
+pytest==6.2.4
 python-dateutil==2.8.1
 python-dotenv==0.15.0
 python-editor==1.0.4
--- a/utils.sh
+++ b/utils.sh
@@ -4,14 +4,20 @@ then
     then
         docker exec -it planet-read_py-backend_1 pytest
     else
-        # TODO improvement - parse from backend/ path location
-        # right now, has to be from app/...
-        docker exec -it planet-read_py-backend_1 pytest $2
+        if [[ "$3" = "" ]]
+        then
+            # TODO improvement - parse from backend/ path location
+            # right now, has to be from app/...
+            docker exec -it planet-read_py-backend_1 pytest $2
+        else
+            # super detailed, match a test name pattern
+            docker exec -it planet-read_py-backend_1 pytest $2 -k "$3" -vv
+        fi
     fi
 elif [[ "$1" = "enter-db" ]]
 then
     ROOT_PASSWORD=$(grep MYSQL_ROOT_PASSWORD .env | cut -d '=' -f2)
-    docker exec -it planet-read_db_1 mysql -u $ROOT_PASSWORD -proot
+    docker exec -it planet-read_db_1 mysql -u root -p$ROOT_PASSWORD
 else
     echo "Invalid command."
 fi
--- a/utils.sh
+++ b/utils.sh
@@ -1,0 +1,17 @@
+if [[ "$1" = "pytest" ]]
+then
+    if [[ "$2" = "" ]]
+    then
+        docker exec -it planet-read_py-backend_1 pytest
+    else
+        # TODO improvement - parse from backend/ path location
+        # right now, has to be from app/...
+        docker exec -it planet-read_py-backend_1 pytest $2
+    fi
+elif [[ "$1" = "enter-db" ]]
+then
+    ROOT_PASSWORD=$(grep MYSQL_ROOT_PASSWORD .env | cut -d '=' -f2)
+    docker exec -it planet-read_db_1 mysql -u $ROOT_PASSWORD -proot
+else
+    echo "Invalid command."
+fi
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 **/build
 **/venv
 **/__pycache__
+**/.pytest_cache
 **/*.log
 **/firebaseServiceAccount.json
 **/.DS_Store
--- a/README.md
+++ b/README.md
@@ -54,13 +54,13 @@ Backend test run using the `pytest` framework on a distinct test database. You m
 
 If you want to run any other pytest commands, run them manually with:
 ```
-docker exec <backend-contianer-id> /bin/bash <pytest command>
+docker exec -it planet-read_py-backend_1 /bin/bash <pytest command>
 ```
 
 # Lint
 Frontend has on-save linting. To lint the backend:
 ```
-docker exec <backend-container-id> /bin/bash -c "black . && isort --profile black ."
+docker exec -it planet-read_py-backend_1 /bin/bash -c "black . && isort --profile black ."
 ```
 
 Follow the [getting started](https://uwblueprint.github.io/starter-code-v2/docs/getting-started) for more details, especially if you desire to use your own firebase and gcp projects.
--- a/backend/python/app/__init__.py
+++ b/backend/python/app/__init__.py
@@ -50,7 +50,9 @@ def create_app(config_name):
     ] = "mysql://{username}:{password}@{host}:3306/{db}".format(
         username=os.getenv("MYSQL_USER"),
         password=os.getenv("MYSQL_PASSWORD"),
-        host=os.getenv("TEST_DB_HOST") if config_name == "testing" else os.getenv("DB_HOST"),
+        host=os.getenv("TEST_DB_HOST")
+        if config_name == "testing"
+        else os.getenv("DB_HOST"),
         db=os.getenv("MYSQL_DATABASE"),
     )
 
--- a/backend/python/app/__init__.py
+++ b/backend/python/app/__init__.py
@@ -50,7 +50,7 @@ def create_app(config_name):
     ] = "mysql://{username}:{password}@{host}:3306/{db}".format(
         username=os.getenv("MYSQL_USER"),
         password=os.getenv("MYSQL_PASSWORD"),
-        host=os.getenv("DB_HOST"),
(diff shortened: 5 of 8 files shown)


backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,164 @@
+import re
+
+from ....models.story import Story
+
+
+def test_create_story(app, db, client):
```
e-wai, line 6: Lots of concerns here:

- Well this test fails for a type mismatch. Is this expected?
```
At index 0 diff: {'id': 1, 'story_id': '1', 'line_index': 0, 'cont
ent': 'line 1'} != {'id': 1, 'story_id': 1, 'line_index': 0, 'content': 'line 1'
```
- This is feels kinda hacky, particularly how I'm trying to work around the snake case conversion, and nesting of the contents array. 
- What do we actually want to test here? I've basically asserted that the db entry and response against expected. Is that right?

backend/python/app/tests/graphql/queries/test_story_query.py
```diff
@@ -0,0 +1,81 @@
...
+    db, client
+):
+    pass
+
+
+def test_stories_available_for_review_user_not_approved(db, client):
+    pass
+
+
+def test_not_authenticated_for_endpoint(db, client):
```
gaoxk, line 80: hmm perhaps there should be a test file for gql middleware rather than this test in `test_story_query.py`. It can test a gql request using `@require_authorization_by_role_gql` . Thoughts?

backend/python/app/tests/services/test_story_service.py
```diff
@@ -0,0 +1,124 @@
...
+    # I leave it here as a reference for how to create objects
+    obj = Story(title="title", description="description", youtube_link="", level=1)
+    db.session.add(obj)
+    db.session.commit()
+    fetched = Story.query.all()
+    assert len(fetched) == 1
+    assert fetched[0].title == "title"
+
+
+def assert_story_equals_model(story_response, story_model):
```
gaoxk, line 16: would prefer if this is at the top of the file, bottom of the file, or in a util file.
could also be a function written into the Story model! like to_dict

utils.sh
```diff
@@ -0,0 +1,23 @@
...
+            docker exec -it planet-read_py-backend_1 pytest $2
+        else
+            # super detailed, match a test name pattern
+            docker exec -it planet-read_py-backend_1 pytest $2 -k "$3" -vv
+        fi
+    fi
+elif [[ "$1" = "enter-db" ]]
+then
+    ROOT_PASSWORD=$(grep MYSQL_ROOT_PASSWORD .env | cut -d '=' -f2)
+    docker exec -it planet-read_db_1 mysql -u $ROOT_PASSWORD -proot
```
gaoxk, line 20: nit: this would actually be `docker exec -it planet-read_db_1 mysql -u root -p$ROOT_PASSWORD `
It works in our current environment since the root password is `root`.

backend/python/app/tests/services/test_story_service.py
```diff
@@ -0,0 +1,124 @@
...
+
+def test_create_translation():
+    pass
+
+
+def test_create_translation_raises_error_if_story_translation_commit_fails_and_nothing_saved():
+    pass
+
+
+def test_create_translation_raises_error_if_translation_conteny_commit_fails_and_nothing_saved():
```
gaoxk, line 64: nit: typo in content

gaoxk (changes requested):
**good shit 🆗 👌 thats the GOOD shit 💯 🥵 🥵 goODD 👍** 

A few patterns I see in naming:
- `test_..._raises_error_if_...`
- `test_..._raises_exception_for_...`
- `test_..._raises_exception_if_...`
- `test_..._raises_exception_when_...`
- `test_stories_available_for_review_invalid_language_string` 
  - this is basically `test_..._erroneous-situation`

In terms of exception vs error, python raises exceptions. Exception is a specific python term while error is more generically describing that the user was returned something they did not expect. Personally, I prefer `test_..._situation` as it takes focus off whether or not the result is erroneous and puts focus on the circumstances of the test. Thoughts?

Also, I stan the `app/tests` 👍 
Also, regarding graphene snapshotting - I like the idea of skipping the response checking misery, but we would still need to check the database to ensure the model exists. Base on what you've seen [in the docs](https://docs.graphene-python.org/en/latest/testing/#snapshot-testing) and your pytest journey, is there an easy way to achieve that?

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+    """
+    )
+
+    assert dict(result["data"]) == {
+        "createStory": {
+            "ok": True,
+            "story": {
+                "id": 1,
```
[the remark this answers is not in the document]
e-wai: Think so? I can change the scope on the db fixture if we don't want that but I think it drops and recreates after each test. Though, to be honest, it may speed up our tests a lot if we don't drop/recreate so much

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+    test_db_obj = (
+        Story.query.order_by(Story.id.desc())
+        .first()
+        .to_dict(include_relationships=True)
+    )
+    # will want this somewhere globally i assume
+    pattern = re.compile(r"(?<!^)(?=[A-Z])")
+
+    for field_name, value in result["data"]["createStory"]["story"].items():
+        if field_name == "contents":
```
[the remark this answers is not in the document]
e-wai: Hm what do you mean?

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+
+def test_create_story_translation(db, client):
+    pass
+
+
+def test_create_story_translation_exception(db, client):
+    pass
+
+
+def test_create_story_translation_translation_already_created(db, client):
```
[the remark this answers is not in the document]
e-wai: Yep

backend/python/app/tests/graphql/queries/test_story_query.py
```diff
@@ -0,0 +1,81 @@
+def test_stories(app, db, client):
```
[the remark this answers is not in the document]
e-wai: Does it really test anything different? It's the same case logic-wise

backend/python/app/tests/graphql/queries/test_story_query.py
e-wai: Yeah I like this, can we eliminate all permissions tests then?

backend/python/app/tests/services/test_story_service.py
e-wai: Think this is a similar conversation to the test_story_mutation tests and how we want to test models? But also yes, how do we want to handle helpers?

backend/python/requirements.txt
```diff
@@ -50,6 +50,7 @@ pyasn1-modules==0.2.8
 pycparser==2.20
 pymongo==3.11.3
```
e-wai, line 51: do we ever use this?

backend/python/app/tests/graphql/mutations/test_story_mutation.py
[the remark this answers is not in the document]
gaoxk: oh my bad, mis interepreted something

backend/python/app/tests/graphql/queries/test_story_query.py
gaoxk: > can we eliminate all permissions tests
All that use fixtures, yes. I think some mutations like assignReviewer throw an error if "permissions" arent met (ie user is not qualified to be reviewer)

backend/python/requirements.txt
gaoxk: good catch - looks like its for mongodb, so no

gaoxk (approved):

gaoxk:
woohoo! 🚢 🚢 🚢

PR #65, opened 2021-08-13
Create commentsByStoryTranslation query

gaoxk:
## Implementation description
* add get_comments_by_story_translation to comment service
* add resolve_comments_by_story_translation

## Steps to test
1. run seed script
2. try [this query](http://localhost:5000/graphql?query=%7B%0A%20%20commentsByStoryTranslation(storyTranslationId%3A13)%7B%0A%20%20%20%20id%0A%20%20%20%20commentIndex%0A%20%20%20%20content%0A%20%20%20%20storyTranslationContentId%0A%20%20%7D%0A%7D) out 

* thoughts on the sqlalchemy query?

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/comment_service.py
+++ b/backend/python/app/services/implementations/comment_service.py
@@ -47,6 +47,25 @@ class CommentService(ICommentService):
 
         return new_comment
 
+    def get_comments_by_story_translation(self, story_translation_id):
+        try:
+            comments = (
+                Comment.query.join(Comment.story_translation_content, aliased=True)
+                .filter_by(story_translation_id=story_translation_id)
+                .all()
+            )
+            print(comments)
+        except Exception as error:
+            reason = getattr(error, "message", None)
+            self.logger.error(
+                "Failed to update comment. Reason = {reason}".format(
+                    reason=(reason if reason else str(error))
+                )
+            )
+            raise error
+
+        return comments
+
     def update_comment(self, updated_comment):
         try:
             comment = Comment.query.filter_by(id=updated_comment.id).first()
--- a/backend/python/app/services/implementations/comment_service.py
+++ b/backend/python/app/services/implementations/comment_service.py
@@ -54,14 +54,8 @@ class CommentService(ICommentService):
                 .filter_by(story_translation_id=story_translation_id)
                 .all()
             )
-            print(comments)
         except Exception as error:
-            reason = getattr(error, "message", None)
-            self.logger.error(
-                "Failed to update comment. Reason = {reason}".format(
-                    reason=(reason if reason else str(error))
-                )
-            )
+            self.logger.error(str(error))
             raise error
 
         return comments
--- a/backend/python/app/services/interfaces/comment_service.py
+++ b/backend/python/app/services/interfaces/comment_service.py
@@ -18,6 +18,17 @@ class ICommentService(ABC):
         pass
 
     @abstractmethod
+    def get_comments_by_story_translation(self, story_translation_id):
+        """Return a list of Comments from a given StoryTranslation
+
+        :param story_translation_id: story translation id
+        :return: list of comments
+        :rtype: list of CommentResponseDTO
+        :raises Exception: if id of comment to be updated does not exist
+        """
+        pass
+
+    @abstractmethod
     def update_comment(self, updated_comment):
         """Update an existing Comment object
 
--- a/backend/python/app/graphql/queries/comment_query.py
+++ b/backend/python/app/graphql/queries/comment_query.py
@@ -1,0 +1,5 @@
+from ..service import services
+
+
+def resolve_comments_by_story_translation(root, info, story_translation_id):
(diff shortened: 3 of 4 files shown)


backend/python/app/services/implementations/comment_service.py
```diff
@@ -47,6 +47,25 @@ def create_comment(self, comment):
...
         return new_comment
 
+    def get_comments_by_story_translation(self, story_translation_id):
+        try:
+            comments = (
+                Comment.query.join(Comment.story_translation_content, aliased=True)
+                .filter_by(story_translation_id=story_translation_id)
+                .all()
+            )
+            print(comments)
```
e-wai, line 57: delete?

backend/python/app/services/implementations/comment_service.py
```diff
@@ -47,6 +47,25 @@ def create_comment(self, comment):
...
+        except Exception as error:
+            reason = getattr(error, "message", None)
+            self.logger.error(
+                "Failed to update comment. Reason = {reason}".format(
```
e-wai, line 61: comment inaccurate

backend/python/app/services/interfaces/comment_service.py
```diff
@@ -17,6 +17,17 @@ def create_comment(self, comment):
...
         pass
 
+    @abstractmethod
+    def get_comments_by_story_translation(self, story_translation_id):
+        """Return a list of Comments from a given StoryTranslation
+
+        :param story_translation_id: story translation id
+        :return: list of comments
+        :rtype: list of CommentResponseDTO
+        :raises Exception: if id of comment to be updated does not exist
```
e-wai, line 27: could we get exceptions if something about the database fails?

e-wai (changes requested):
Query itself looks good, I'm curious what the use case for it is:

Right now, the translation page query is
https://github.com/uwblueprint/planet-read/blob/1b81691de3b9e0aa379b552d8c0e9f5b1e7f5e5d/frontend/src/components/pages/TranslationPage.tsx#L29-L47

and it looks like with this, the query becomes
```
query { 
  storyById(id: ${storyId}) { ... } 
  storyTranslationById(id: ${storyTranslationId}) { 
     translationContents { 
         id 
         lineIndex 
         content: translationContent 
     }, 
     numTranslatedLines 
  } 
  comments { ... }
} 
```

Could you not just load it on the story translation itself? It would prevent having to do another read  of the db I think, unless it's more inefficient because of how GraphQL resolves fields? ie. Why not
```
query { 
  storyById(id: ${storyId}) { ... } 
  storyTranslationById(id: ${storyTranslationId}) { 
     translationContents { 
         id 
         lineIndex 
         content: translationContent 
     }, 
     numTranslatedLines,
     comments { ... }
  } 
} 
```

backend/python/app/services/interfaces/comment_service.py
gaoxk: yep, all service function would throw exceptions in the event of a database fails (though we have discussed that not all of them are gracefully handeled)

gaoxk:
Why not
```
query { 
  storyById(id: ${storyId}) { ... } 
  storyTranslationById(id: ${storyTranslationId}) { 
     translationContents { 
         id 
         lineIndex 
         content: translationContent 
     }, 
     numTranslatedLines,
     comments { ... }
  } 
} 
```
I've been starting to get a feeling that our queries are getting thicc. Rather than [batching](https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/), so far we've taken to fetching translationcontents and storycontents in queries like storyById and storyTranslationById. I'm hesitant to continue to add to the chains as we might start getting 4 layers thick (story->storyTranslation->storyTranslationContent->comment). We could add a foriegn key on comments with storytranslation and get only 3 layers thick (story->storyTranslation->comment). Based on how #50 is going, we might need to add an additional database call to return comments together with storyTranslation or story if we took a similar route.
Comments side bar might not always need to be opened, especially in the early translation phases for the translator. Leaving the fetch for comments up to the comment side bar component follows some best practice guidelines from apollo, and also gives us freedom to mess around with filter options that shaahana wants
`In the large majority of cases, a query such as the following should be divided into multiple queries that are distributed among the appropriate components`
https://www.apollographql.com/docs/react/data/operation-best-practices/

e-wai (approved):
> Comments side bar might not always need to be opened, especially in the early translation phases for the translator. Leaving the fetch for comments up to the comment side bar component follows some best practice guidelines from apollo, and also gives us freedom to mess around with filter options that shaahana wants

Ooh ok that's interesting. Sounds good to me then.
(Also, it isn't immediately apparent in the designs that the comments bar is collapsible, unless you mean between the translate and review pages)

gaoxk:
yeahh, shaahana is planning on making the toggling more clear. The reviewing mocks has a little translator<>reviewer toggle at the top right

gaoxk:
TFTR! 🚢

PR #72, opened 2021-08-19
Add test skeletons for story service, query, mutation, comment service

e-wai:
Not sure what happened with the other one?

## Notion ticket link

[Add pytest](https://www.notion.so/uwblueprintexecs/Add-pytest-78637b82f2d9450eab9ef72dcf011290)
- This is part 2, find part 1 here: https://github.com/uwblueprint/planet-read/pull/63

## Usage
Test files are found by the following:
- Tests are in a `tests/` directory
- The test file is named `test_<something>.py`
- Each test is defined in a function whose prefix is `test_`:
```python
def test_<describe_the_test>():
  pass
```
The available fixtures currently are `app`, `db`, `client`, and `services`. If you pass the name of the fixture as a parameter of the test, it will execute fixture code before running your test and return or yield to you.
```python
def test_<describe_the_test>(app):
  result = app.run_something(True)
  assert(result)
```

## Implementation description
* Trivia no. 2: ran into a ton of problems with parent packages not being found (run https://github.com/uwblueprint/planet-read/pull/53/ to observe the 😭) and restructuring made it clear where I could insert `__init__.py` files and eventually worked

## Steps to test
1. Rebuild (`docker-compose down -v`, `docker-compose up --build`) and recreate db 
2. Run the tests using `./utils.sh pytest`
3. Try out the other pytest command options: for a specific file, and for a file + name pattern match. Run it manually with `docker exec -it <backend-container-id> /bin/bash <pytest command>`
4. Verify that the tests db and general db refer to different instances: Make a change (whether via backend or frontend) to one of the tables, run the tests (which will drop and recreate everything), and check that your change has persisted. 

* Test structure: I think there are two schools of thoughts here, (1) that tests should live in a distinct `app/tests` directory that mirrors the file structure of the rest of the app, or (2) that tests should live right next to the actual code. I opted for (1) because it made it easy to add `__init__.py` files that made it work and I don't think imports look too, too bad. 
* Fixture placement: all our fixtures are in a single location right now, `conftest.py`. I think we should split this up as we go. Similar thought for helpers - we'll need somewhere to put them that isn't just one long file
* The one mutation test that was implemented, `test_story_mutation.py` does a lot of verifying, checking both the expected response type and that the expected changes were made to the db for each field. This causes some problems: there are some weird type mismatches and one is snake case vs. other in camel case, which with nesting, is a pain to sort out. Is this the right approach/extensiveness at which we want to test?
* Somewhat related, Graphene offers an option for snapshot testing graphene (see [docs](https://docs.graphene-python.org/en/latest/testing/#snapshot-testing)). I'm not sold on using it, but open to feedback -- it is much more concise, but somewhat hides our result from us.
* Read through the tests cases. This PR isn't focused on extensive test coverage (so naming isn't consistent, probably missing cases, and I definitely use error vs. exception interchangeably) and will grow as we implement more, but let me know if it all makes sense.

## Checklist


backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
+import re
+
+from ....models.story import Story
+
+
```
e-wai, line 5: Lots of concerns here:

- Well this test fails for a type mismatch.
```
At index 0 diff: {'id': 1, 'story_id': '1', 'line_index': 0, 'cont
ent': 'line 1'} != {'id': 1, 'story_id': 1, 'line_index': 0, 'content': 'line 1'
```
Conversely, when I make the expected type into char, the dict conversion wants `line_index` as an integer
```
 {'createStory': {'ok': True, 'story': {'contents': [{'content': 'line 1', 'id': 1, 'lineIndex': 0, 'storyId': '1'}, {'...'id': 3, 'lineIndex': 2, 'storyId': '1'}]
, 'description': 'this should explain things', 'id': 1, 'level': 10001, ...}}} != {'createStory': {'ok': True, 'story': {'contents': [{'content': 'line 1', 'id': 1, 'lineIn
dex': '0', 'storyId': '1'}, ...d': 3, 'lineIndex': '2', 'storyId': '1'}], 'description': 'this should explain things', 'id': 1, 'level': 10001, ...}}}
```
- This is feels kinda hacky, particularly how I'm trying to work around the snake case conversion, and nesting of the contents array. 
- What do we actually want to test here? I've basically asserted that the db entry and response against expected. Is that right?

e-wai: ### Reply from @gaoxk:

```
At index 0 diff: {'id': 1, 'story_id': '1', 'line_index': 0, 'cont
ent': 'line 1'} != {'id': 1, 'story_id': 1, 'line_index': 0, 'content': 'line 1'
```
This is happening because `value` and `test_db_obj[to_snake_case(pattern, field_name)]` are different dict objects in python's memory. [Something like this will work.](https://stackoverflow.com/questions/4527942/comparing-two-dictionaries-and-checking-how-many-key-value-pairs-are-equal)

>This is feels kinda hacky, particularly how I'm trying to work around the snake case conversion, and nesting of the contents array.

The snake case conversion is quite unfortunate. This test also has some repetition; story data and contents are explicitly written out in the gql request and the response test. Declaring the expected model first may alleviate some pains:
```
new_story = Story(title="title test", description="this should explain things", ...)
result = client.execute(
        """
         mutation CreateStory($storyData: ...){
            createStory(
                contents: $contents,
...
       variables={storyData:{new_story.title, ...}...},
)
...
assert dict(result["data"]) == {
        "createStory": {
            "ok": True,
            "story": new_story.to_dict(camel_case=True, include_relationships=True)
....
assert test_db_obj.title == new_story.title
# could be replaced with a DIY generic attribute checker, sqlalchemy doesn't have a built in
# could be replaced with assert_story_equals_model from test_story_service.py
...
```

> What do we actually want to test here? I've basically asserted that the db entry and response against expected. Is that right?

yep! The essential concept looks good to me

backend/python/app/tests/services/test_story_service.py
```diff
@@ -0,0 +1,124 @@
...
+    # I leave it here as a reference for how to create objects
+    obj = Story(title="title", description="description", youtube_link="", level=1)
+    db.session.add(obj)
+    db.session.commit()
+    fetched = Story.query.all()
+    assert len(fetched) == 1
+    assert fetched[0].title == "title"
+
+
+def assert_story_equals_model(story_response, story_model):
```
e-wai, line 16: @gaoxk 
> would prefer if this is at the top of the file, bottom of the file, or in a util file.
> could also be a function written into the Story model! like to_dict

@e-wai 
> Think this is a similar conversation to the test_story_mutation tests and how we want to test models? But also yes, how do we want to handle helpers?

backend/python/app/tests/graphql/queries/test_story_query.py
```diff
@@ -0,0 +1,81 @@
...
+    db, client
+):
+    pass
+
+
+def test_stories_available_for_review_user_not_approved(db, client):
+    pass
+
+
+def test_not_authenticated_for_endpoint(db, client):
```
e-wai, line 80: @gaoxk 

> hmm perhaps there should be a test file for gql middleware rather than this test in test_story_query.py. It can test a gql request using @require_authorization_by_role_gql . Thoughts?

@e-wai 
> Yeah I like this, can we eliminate all permissions tests then?

@gaoxk 
> All that use fixtures, yes. I think some mutations like assignReviewer throw an error if "permissions" arent met (ie user is not qualified to be reviewer)

backend/python/app/tests/graphql/queries/test_story_query.py
```diff
@@ -0,0 +1,81 @@
+def test_stories(app, db, client):
```
e-wai, line 1: @gaoxk 
> nit: maybe this should be split into test_stories_empty and test_stories

@e-wai 
> Does it really test anything different? It's the same case logic-wise

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+def test_create_story_exception(db, client):
+    pass
+
+
+def test_create_story_translation(db, client):
+    pass
+
+
+def test_create_story_translation_exception(db, client):
```
e-wai, line 86: @gaoxk 
> Not a fan of the test_xyzflow_exception - the goal is to test all situations, some of which result in exceptions. Test like test_xyzflow_exception imply tests should be written with a specific outcome in mind, rather than a specific situation

> I do see the appeal of ensuring every service function has proper exception handling, as you discovered some services lacked. What are your thoughts? In the generic exception scenario, the database or client likely has to throw an exception - is there an easy way to cause an erroneous test with the fixture db or client?

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+    pass
+
+
+def test_create_story_translation_translation_already_created(db, client):
```
e-wai, line 90: @gaoxk 
> to confirm: this is testing the case where the story translation for a specific language has already been created?

@e-wai 
> yep

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+    """
+    )
+
+    assert dict(result["data"]) == {
+        "createStory": {
+            "ok": True,
+            "story": {
+                "id": 1,
```
e-wai, line 45: @gaoxk 
> Is the database cleared after each test? Will the id always be 1?

@e-wai 
> Think so? I can change the scope on the db fixture if we don't want that but I think it drops and recreates after each test (function scope). Though, to be honest, it may speed up our tests a lot if we don't drop/recreate so much

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,163 @@
...
+                "contents": [
+                    {"id": 1, "storyId": "1", "lineIndex": 0, "content": "line 1"},
+                    {"id": 2, "storyId": "1", "lineIndex": 1, "content": "line 2"},
+                    {"id": 3, "storyId": "1", "lineIndex": 2, "content": "line 3"},
+                ],
+            },
+        }
+    }
+
+    test_db_obj = (
```
e-wai, line 60: @gaoxk 
> Perhaps the SQLalchemy object should be found like
> Story.query.filter(id=returned_id).to_dict(include_relationships=True)

e-wai (commented):
Copied over all review comments left unresolved so far from the initial PR (#63)

### @gaoxk overall comments

**good shit 🆗 👌 thats the GOOD shit 💯 🥵 🥵 goODD 👍** 

A few patterns I see in naming:
- `test_..._raises_error_if_...`
- `test_..._raises_exception_for_...`
- `test_..._raises_exception_if_...`
- `test_..._raises_exception_when_...`
- `test_stories_available_for_review_invalid_language_string` 
  - this is basically `test_..._erroneous-situation`

In terms of exception vs error, python raises exceptions. Exception is a specific python term while error is more generically describing that the user was returned something they did not expect. Personally, I prefer `test_..._situation` as it takes focus off whether or not the result is erroneous and puts focus on the circumstances of the test. Thoughts?

Also, I stan the `app/tests` 👍 
Also, regarding graphene snapshotting - I like the idea of skipping the response checking misery, but we would still need to check the database to ensure the model exists. Base on what you've seen [in the docs](https://docs.graphene-python.org/en/latest/testing/#snapshot-testing) and your pytest journey, is there an easy way to achieve that?

backend/python/app/tests/graphql/queries/test_story_query.py
gaoxk: I think we ultimately would want different tests for an empty return/empty database state vs a filled database

backend/python/app/tests/services/test_story_service.py
gaoxk: by helpers do you mean functions like `assert_story_equals_model`? something like a file `app/tests/utils.py` could work, or the functions could be baked into models

e-wai: Created a separate directory for helpers, `app/tests/helpers` for eg. `app/tests/helpers/story_helper.py`. Do you think there will be more general helpers that wouldn't belong to one of these?

backend/python/app/tests/graphql/mutations/test_story_mutation.py
e-wai: Much better, went 124.30s -> 13.22s

backend/python/app/tests/graphql/queries/test_story_query.py
e-wai: updated with a new `test_auth.py`, let me know what you think

backend/python/app/tests/graphql/mutations/test_story_mutation.py
e-wai: From discussions -> keep snapshots ([docs](https://docs.graphene-python.org/en/latest/testing/#snapshot-testing)) as an investigation in future PR. Not sure about it, so avoid having to rebuild

backend/python/app/tests/graphql/mutations/test_story_mutation.py
e-wai: I'm thinking that we can structure as:
- test_success
- test_known_failure
- test_known_failure
- ...
- test_unexpected_exception (for handling)

Need to look into it more, but I'm hoping we can use the `mock` package to throw an exception

backend/python/app/tests/graphql/queries/test_story_query.py
e-wai: Note to self → need to guarantee data in database. Empty database was hiding problems in formatting

backend/python/app/tests/graphql/mutations/test_story_mutation.py
```diff
@@ -0,0 +1,161 @@
...
+        },
+    )
+
+    returned_dict = result["data"]["createStory"]
+    story_db_id = returned_dict["story"]["id"]
+    test_db_story = Story.query.get(story_db_id).to_dict(include_relationships=False)
+    test_db_story_contents = StoryContent.query.filter_by(story_id=story_db_id).all()
+
+    assert returned_dict["ok"]
+    assert returned_dict["story"]["title"] == new_story.title
```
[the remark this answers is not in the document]
e-wai: Changed up so it's used for story model, but didn't find an intuitive way to include content without a double iteration

e-wai:
Outstanding discussion items:
- deleted all db/generic exception failures - discussion resolved? 
  - snapshotting and mocking libraries investigation can proceed later if find the need
- see the `require_story_in_db` func guaranteeing data using a helper. Is a fixture preferred? I find this easier to read (also, perhaps more straightforward to implement)
- running tests in CircleCI, parallelizing across CPUs
- what `pytest` commands in utils do we want? apart from existing ones
- from #internal-tools-help, decided to test auth through integration/e2e tests
- please double-check the type mismatch note - is this expected that the foreign id is returned as a str?

backend/python/app/tests/helpers/story_helpers.py
```diff
@@ -0,0 +1,23 @@
...
+        assert story_response["youtubeLink"] == story_model.youtube_link
+        # TODO: expected translatedLanguages?
+        assert story_response["translatedLanguages"] == None
+    else:
+        assert story_response["youtube_link"] == story_model.youtube_link
+        # TODO: expected translated_languages?
+        assert story_response["translated_languages"] == None
+
+
+def require_story_in_db(db):
```
gaoxk, line 19: nit: add a quick comment explaining what this function does

backend/python/app/tests/helpers/story_helpers.py
```diff
@@ -0,0 +1,23 @@
...
+    if Story.query.count() == 0:
+        obj = Story(title="title", description="description", youtube_link="", level=1)
```
gaoxk, line 21: nit: some real data would be nice - feel free to take something from the seeding scripts

gaoxk (approved):
> deleted all db/generic exception failures - discussion resolved?

Yee discussion resolved imo

> see the require_story_in_db func

see comment on that file

> running tests in CircleCI, parallelizing across CPUs

I'll make a ticket for this 👍 

> what pytest commands in utils do we want? do the existing ones cover the more common cases?

I'm happy with the existing cases :) 

Excellent work! feel free to merge after resolving nits

backend/python/app/tests/helpers/story_helpers.py
e-wai: renamed to a more descriptive `guarantee_story_table_not_empty`

PR #80, opened 2021-08-30
Add database migrations to repository

jennchenn:
## Notion ticket link

[Add database migrations to repository](https://www.notion.so/uwblueprintexecs/Add-database-migrations-to-repository-9dbf580ccd994b90aa613e97a26fc7f2)

## Implementation description
* Generate initial database migration using Flask-Migrate

## Steps to test
1. Add the seed_db.sh `AUTH_ID`s to your root `.env` file
2. Rebuild containers: `docker-compose down --volumes` and `docker-compose up --build`
3. Run `docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db upgrade"`
4. Verify that the database tables are generated by checking the database itself
5. Run `docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db downgrade"` and verify the tables are dropped: `docker exec -it planet-read_db_1 mysql -u root -proot -e "USE planet-read; SHOW TABLES;"` 

* `backend/python/migrations/versions/3f6fff20884c_initial_migration.py` (most of the other files in `backend/python/migrations/` were auto-generated/modified based on Starter Code)
~~* Is there a better place to put the seeded data than in the initial migration itself? This seemed to be the pattern that turned up the most when researching. A reference to the create table object is used in the `op.bulk_insert()` command, but an alternate command can be used if we want to separate it out (i.e. `op.execute()`. However, I believe alembic revisions are linear, so if we create a separate migration specifically for test data and then decide to remove it once we release to production, we will need to make sure we update the revision heads of the migrations to ensure the `flask db upgrade` command runs.~~
* The `AUTH_ID_[1-7]`s that were used for the seed_db.sh would need to be added to the vault

## Checklist


backend/python/Dockerfile
```diff
@@ -6,7 +6,6 @@ COPY requirements.txt ./
 RUN pip install -r requirements.txt
 
 COPY . ./app
-VOLUME file_uploads:/var/lib/file_uploads
```
jennchenn, line 9: i think this double declaration in here and the compose file might have caused the file volume OCI runtime error. however i got the error pretty inconsistently before making this change so not sure if it's just coincidence that i haven't seen it again. removing it doesn't seem to cause any issues with running the containers though

backend/python/migrations/versions/3f6fff20884c_initial_migration.py
```diff
@@ -0,0 +1,523 @@
...
+    op.create_index(
+        op.f("ix_comments_story_translation_content_id"),
+        "comments",
+        ["story_translation_content_id"],
+        unique=False,
+    )
+    ### end Alembic commands ###
+
+    # seed database
+    op.bulk_insert(
```
gaoxk, line 171: image
woops, didnt realize seeding would be included in this pr, closing #81. One modification I'll ask for: can you copy the descriptions and youtube links for stories from #81?

README.md
```diff
@@ -63,6 +63,25 @@ Frontend has on-save linting. To lint the backend:
...
+```
+docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db downgrade"
+``` 
+
+
+A new migration will need to be generated when database changes are made. Import any new tables into [backend/python/app/models/\_\_init__.py](backend/python/app/models/__init__.py) and run: 
+```
+docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db migrate -m '<description of your migration>'"
+```
+Ensure that a new revision file is created in the directory [backend/python/migrations/versions](backend/python/migrations/versions).
```
gaoxk, line 81: nit: I'd add that generally these auto generated files should be left untouched (initial db seeding is an exception)

gaoxk (changes requested):
> if we create a separate migration specifically for test data and then decide to remove it once we release to production, we will need to make sure we update the revision heads of the migrations to ensure the flask db upgrade command runs.

Can you describe how we would go about swapping initial seed/test data migrations (would it actually just be a alembic script?) for production vs development environments? 
To distinguish test data vs seed data: test data is what we currently have going on - it's fake data that helps us test and develop. seed data is data that must be seeded to make the application useable (eg, initial admin accounts). inserting test data should not happen on prod, but inserting seed data should.
A very hacky way to work with the separate migrations is to make sure the test data migration (current seed_db results) dont get copied to the client's server. Curious what your thoughts are.

gaoxk:
((also, tested it and it worked!))

backend/python/migrations/versions/3f6fff20884c_initial_migration.py
jennchenn: Removed seeding; adding to this PR instead: https://github.com/uwblueprint/planet-read/pull/83

README.md
```diff
@@ -63,6 +63,25 @@ Frontend has on-save linting. To lint the backend:
...
+```
+docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db downgrade"
+``` 
+
+
+A new migration will need to be generated when database changes are made. Import any new tables into [backend/python/app/models/\_\_init__.py](backend/python/app/models/__init__.py) and run: 
+```
+docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db migrate -m '<description of your migration>'"
+```
+Ensure that a new revision file is created in the directory [backend/python/migrations/versions](backend/python/migrations/versions). Do _not_ change the alembic revision/identifiers. Generally these auto-generated revision files will encompass all schema changes, and thus do not need to be modified!
```
gaoxk, line 81: nit: bold **Do not** rather than italicize just not

gaoxk (approved):
noooooooooooooooooice!

PR #82, opened 2021-09-01
Implemented debounced undo

ZhangCreations:
## Notion ticket link
[Debounce translation undo redo](https://www.notion.so/uwblueprintexecs/Debounce-translation-undo-redo-12b5178abf5e4120ab34f23e666d190c)

## Implementation description
* The name is a misnomer, since it's not quite debounced, since users might want to have more granularity in undos than what the debounced approach would offer. 
* Instead, the state is stored every second and users can iterate over those in the undo, otherwise all functionality is essentially the same as https://github.com/uwblueprint/planet-read/pull/55.

## Steps to test
1. Open a story
2. Type some things
3. Undo some things
4. Redo some things
5. Try undoing all your changes and redoing all your changes
6. Try undoing after no changes
7. Give yourself a pat on the back for being a cool person :)

* Not sure about how the timings feel, whether or not the exact amount of time should be changed, let me know what you think.

## Checklist

<p align="center">
  
</p>

### DIFF
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { useQuery } from "@apollo/client";
 import { Icon } from "@chakra-ui/icon";
 import { Box, Button, Divider, Flex, IconButton, Text } from "@chakra-ui/react";
@@ -27,8 +27,10 @@ type Content = {
 };
 
 type HistoryStack = {
-  Undo: Array<{ lineIndex: number; content: string }>;
-  Redo: Array<{ lineIndex: number; content: string }>;
+  Undo: Array<StoryLine[]>;
+  Redo: Array<StoryLine[]>;
+  // Undo: Array<{ lineIndex: number; content: string }>;
+  // Redo: Array<{ lineIndex: number; content: string }>;
 };
 
 const TranslationPage = () => {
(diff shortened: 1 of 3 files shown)


frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -89,77 +91,77 @@ const TranslationPage = () => {
     );
   };
 
+  function saveVersionState() {
```
gaoxk, line 94: nit: `function saveVersionState() {` => `const saveVersionState = () => {`

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -25,8 +25,10 @@ type Content = {
 };
 
 type HistoryStack = {
-  Undo: Array<{ lineIndex: number; content: string }>;
-  Redo: Array<{ lineIndex: number; content: string }>;
+  Undo: Array<StoryLine[]>;
+  Redo: Array<StoryLine[]>;
+  // Undo: Array<{ lineIndex: number; content: string }>;
```
gaoxk, line 30: nit: remove comments

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -25,8 +25,10 @@ type Content = {
 };
 
 type HistoryStack = {
-  Undo: Array<{ lineIndex: number; content: string }>;
-  Redo: Array<{ lineIndex: number; content: string }>;
+  Undo: Array<StoryLine[]>;
```
gaoxk, line 28: having a somewhat hard time following the logic -  why switch to an array of `StoryLine[]`?

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -89,77 +91,77 @@ const TranslationPage = () => {
...
+        Undo: [
+          ...deepCopy(versionHistoryStack.Undo),
+          deepCopy(translatedStoryLines),
+        ],
+        Redo: [],
+      });
+    }
+  }
+
+  useEffect(() => {
```
gaoxk, line 113: interesting approach - did you consider extending `Autosave` code/functionality? How did that go over? @e-wai might have some more ideas as she implemented it

gaoxk (changes requested):
Seems a little buggy after doing a full set of undos followed by redos
recording: https://www.loom.com/share/32728125b1b74253a2a3f83e78130090 
skip to :35 for the undo-ing to begin

e-wai (dismissed):
This is more about debouncing than the saving part of it - can we add some sort of indicator to the user of "saving..." vs. "saved" states? @gaoxk

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: What Emma said is exactly correct,  we can't actually know which line the user will change until they actually change it, so when the state is stored, it needs to be all lines.

frontend/src/components/pages/TranslationPage.tsx
ZhangCreations: I did consider that, but ended up not doing that because the debouncing on Autosave doesn't feel intuitive because of the long pause required for the autosave to kick in

ZhangCreations:
No idea why I'm using useRef instead of useState here, just followed the linked guide and it worked so ¯\_(ツ)_/¯. Anyways, the method for keeping track of state has been largely changed so if anyone would like a more in depth explanation how it's being done lemme know.

gaoxk (approved):
Noice! Gunna push some changes myself since Kevin's tenure is done.

Puepis (commented):
I tested it out on my end and I'm running into a case where the text isn't autosaved after undo-ing, not sure what the intended behaviour is supposed to be

recording: https://www.loom.com/share/25a6996d0f0e473f95f601a499080cfd

gaoxk:
💀 urgh. 
I _could_ have tightened up `versionHistoryStack` to be only relevant changed lines like `snapShotLineIndexes`, but to move between undo/redos, that requires saving the undone _and_ redone lines at the time of saving the version, which is a pain in the arse because you *cant* know the future at the time of saving the history stack! so screw that! here's another damn state object!!! 💀

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -67,18 +69,12 @@ const TranslationPage = () => {
...
   };
 
-  const deepCopy = (lines: Object) => {
-    // This is a funky method to make deep copies on objects with primative values
-    // https://javascript.plainenglish.io/how-to-deep-copy-objects-and-arrays-in-javascript-7c911359b089
-    // Should probably go under some util
-    return JSON.parse(JSON.stringify(lines));
-  };
-
   const onChangeTranslationContent = async (
```
Puepis, line 72: how come this is marked as async? it doesn't look like the async is necessary in this function

Puepis (approved):
Tested it again and looks good to me! Just left a comment about one of the async functions

frontend/src/components/pages/TranslationPage.tsx
gaoxk: good catch! does not have to be async. just tested it out too.

gaoxk:
TFTR! 🚀

PR #96, opened 2021-09-15
Make per-line comment button functional

lynn-kim:
## Notion ticket link

[Make per-line comment button functional](https://www.notion.so/uwblueprintexecs/Make-per-line-comment-button-functional-9cb870ccaae44e8eba0f899370ea288f)

## Implementation description
- added flow for adding comments, with the exception of displaying them 
- added state variables at the TranslationPage level, to pass them between TranslationTable and CommentsPanel

Screen Shot 2021-09-14 at 7 48 24 PM
Screen Shot 2021-09-14 at 8 00 04 PM

## Steps to test

1. Open the app frontend and navigate to a story 
2. Click the comment button in the commentspanel (top right) -> comment buttons should appear in each row 
3. Pick a line and click the comment button -> WIPComment component should appear in pannel
4. Add a comment in the WIPComment component  

- Flow as expected?
- Better way to store state? 

## Checklist

### DIFF
--- a/frontend/src/components/pages/ReviewPage.tsx
+++ b/frontend/src/components/pages/ReviewPage.tsx
@@ -42,6 +42,14 @@ const ReviewPage = () => {
   const [title, setTitle] = useState<string>("");
   const [language, setLanguage] = useState<string>("");
 
+  const [isCommenting, setIsCommenting] = useState(false);
+  const [commentLine, setCommentLine] = useState(0);
+
+  const [
+    storyTranslationContentId,
+    setStoryTranslationContentId,
+  ] = useState<number>(-1);
+
   const handleFontSizeChange = (val: string) => {
     setFontSize(val);
   };
@@ -105,6 +113,9 @@ const ReviewPage = () => {
               fontSize={fontSize}
               originalLanguage="English"
               translatedLanguage={convertLanguageTitleCase(language)}
+              isCommenting={isCommenting}
+              setCommentLine={setCommentLine}
+              setStoryTranslationContentId={setStoryTranslationContentId}
             />
           </Flex>
           <Flex margin="20px 30px" justify="flex-start" alignItems="center">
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/ReviewPage.tsx
```diff
@@ -42,6 +42,14 @@ const ReviewPage = () => {
   const [title, setTitle] = useState<string>("");
   const [language, setLanguage] = useState<string>("");
 
+  const [isCommenting, setIsCommenting] = useState(false);
```
gaoxk, line 45: hmmm perhaps instead of managing `isCommenting` and `commentLine`, we can use only `commentLine`. `commentLine == -1` can be equivalent to 'there is no comment in WIP'. Thoughts?

gaoxk: `storyTranslationContentId` could be used in place of isCommenting too but i could be wrong, In either case, the `commentLine` or `storyTranslationContentId` would need to be reset after a work-in-progress comment is created and the WIPComment component is collapsed

frontend/src/components/pages/ReviewPage.tsx
```diff
@@ -42,6 +42,14 @@ const ReviewPage = () => {
...
+  const [commentLine, setCommentLine] = useState(0);
+
+  const [
```
gaoxk, line 48: this state name is a little ambiguous - can it be `commentStoryTranslationContentId` or something similar?

gaoxk (commented):
Love to see it! 🎉 
General /open ended question: Do you have an idea on how to expand the comments panel to handle replying to comments within the comments panel? would it be easy to have a state where multiple WIPComment components are open? (eg a reviewer is the middle of new comment for a line, and in the middle of a comment reply)

lynn-kim:
@gaoxk hmm I didn't consider the case with multiple WIP comments, but would an array of comment objects work? something like:
```
let WIPComments = [
  {
    commentStoryTranslationId: 10,
    content: "This is a comment", 
  },
  {
    commentStoryTranslationId: 6,
    content: "This is a different comment", 
  },
]
```

gaoxk (changes requested):
excellenté!

gaoxk (approved):
🤦 wrong button. noice!

PR #98, opened 2021-09-17
Create updateStoryTranslationStage mutation

Puepis:
## Notion ticket link

[Create updateStoryTranslationStage](https://www.notion.so/uwblueprintexecs/Create-updateStoryTranslationStage-c6bc9384ff07447e84f5a1f24eec627c)

## Implementation description

- created a method in StoryService called `update_story_translation_stage` that updates the story translation's stage to the new stage if the user is authorized to perform the operation
- the mutation resolver pretty much just calls the method
- currently we only care about updating the stage to `TRANSLATE` or `REVIEW`, updating to `PUBLISH` will be handled separately in a different mutation

## Steps to test

1. Go to `localhost:5000/graphql`

2. To test the mutation you will need the `id` of a story_translation as well as its `translator_id`, for testing I used `id: 12` and `translator_id: 3`, which corresponds to user "Kevin Debryune"

3. Run the following mutation:

```
mutation {
  updateStoryTranslationStage(storyTranslationData: {id: 12, stage: TRANSLATE}) {
    ok
  }
}
```

and you should get an error response saying the user is not authorized to perform the update (only reviewers can move the stage to TRANSLATE)

**Note:** You will have to be signed in as the user "Kevin Debryune" because the method uses the access token from the request

4. Run the same mutation as above except with `stage: REVIEW` and you should get an ok response

5. Verify that the story_translation is now in the `REVIEW` stage by running the query

```
query {
  storyTranslationById(id:12) {
    stage
  }
}
```

- clarity and code readability of the implemented `story_service` method

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -1,6 +1,7 @@
 from flask import current_app
 
 from ...graphql.types.story_type import StageEnum, StoryTranslationContentResponseDTO
+from ...middlewares.auth import get_user_id_from_request
 from ...models import db
 from ...models.story import Story
 from ...models.story_content import StoryContent
@@ -8,7 +9,6 @@ from ...models.story_translation import StoryTranslation
 from ...models.story_translation_content import StoryTranslationContent
 from ...models.story_translation_content_status import StoryTranslationContentStatus
 from ..interfaces.story_service import IStoryService
-from ...middlewares.auth import get_user_id_from_request
 
 
 class StoryService(IStoryService):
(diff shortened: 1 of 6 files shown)


backend/python/app/services/implementations/story_service.py
```diff
@@ -267,6 +268,30 @@ def update_story_translation_contents(self, story_translation_contents):
             )
             raise error
 
+    def update_story_translation_stage(self, story_translation_data):
+        story_translation = StoryTranslation.query.filter_by(
```
jennchenn, line 272: should we wrap this in a try except (like the other functions above) so we catch and log exceptions?

jennchenn (commented):
Haven't had a chance to actually run it but just some comments/questions from reading through the code changes!

backend/python/app/services/implementations/story_service.py
```diff
@@ -267,6 +268,30 @@ def update_story_translation_contents(self, story_translation_contents):
...
+            id=story_translation_data["id"]
+        ).first()
+        new_stage = story_translation_data["stage"]
+        user_id = get_user_id_from_request()
```
gaoxk, line 276: urhg. I couldnt get this mutation working while testing from graphql interface Altair. The issue is that `get_user_id_from_request` returns a string because of `auth_service.user_service.get_user_id_by_auth_id` returning a string. We cannot simply just modify `auth_service.user_service.get_user_id_by_auth_id` to return an int as it's used in multiple places and might cause unforseen issues. 
Instead, for now, use `user_id = int(get_user_id_from_request())` and leave a todo to update the line when `get_user_id_from_request` is updated. I'll work on updating `get_user_id_from_request`

gaoxk (changes requested):
nice! Gunna send message to slack about how to test gql when access tokens need to be involved. lmk when this is good for a rereview - we can approve quickly :)

gaoxk (approved):
nice! hot tip: I like to use `git commit  --fixup <commit sha>` for making commits that address pr fixes. The generated commit will squash into the specified commit. It's not suuuper necessary for our repo since we rebase+squash on merge, but can be helpful for the future. Also indicates small fix commits very clearly.
🚢

backend/python/app/services/implementations/story_service.py
jennchenn: do we need to cast the error to string before logging? `self.logger.error(str(error))` (we seem to do this for most of the other functions)

PR #103, opened 2021-09-25
Reflect filter based on two approved languages pipelines

Puepis:
## Notion ticket link

[Reflect filter based on two approved languages pipelines](https://www.notion.so/uwblueprintexecs/reflect-filter-based-on-two-approved-languages-pipelines-f0b7b4a5aa634f8ea515e47701cbf003)

## Implementation description
- updated mutation and auth user type to fetch approvedLanguagesReview
- minor restyling for Filter component to use margins instead of `justifyContent: space-between`
- display "Role required" box only if `approvedLanguagesReview` object is not empty
- display access level options based on approved level, **previously all levels from 1->4 were displayed**
    

## Steps to test

1. Log in using `planetread+carlsagan@uwblueprint.org`, the sidebar shouldn't display the "Role required" section (since Carl Sagan is only approved as a translator)
2. Clear localStorage and log in as `planetread+miroslavklose@uwblueprint.org`, the sidebar should display the "Translator/Reviewer" selection
3. Switch back and forth between the roles and access levels and verify that the proper languages are shown

- making sure that switching between translator/reviewer shows the appropriate access levels and the correct ones are selected
- dynamically displaying the level options leads to an edge case where switching from a language with a higher level to one with a lower level doesn't change the UI appropriately (will need to refactor `ButtonRadioGroup` in a different ticket)

## Checklist

### DIFF
--- a/frontend/src/components/homepage/Filter.tsx
+++ b/frontend/src/components/homepage/Filter.tsx
@@ -34,27 +34,37 @@ const Filter = ({
   setIsTranslator,
   isDisabled = false,
 }: FilterProps) => {
+  const approvedLanguages = role
+    ? approvedLanguagesTranslation
+    : approvedLanguagesReview;
+
+  // handles edge case of switching from higher approved level to lower level
+  const handleLevelDecrease = (newLevel: number) => {
+    if (level > newLevel) {
+      setLevel(newLevel);
+    }
+  };
+
   const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     setLanguage(event.target.value);
+    handleLevelDecrease(approvedLanguages[event.target.value]);
   };
+
   const handleRoleChange = (nextRole: string) => {
-    const approvedLanguages =
+    const newApprovedLanguages =
       nextRole === "Translator"
         ? approvedLanguagesTranslation
         : approvedLanguagesReview;
-    const newLanguage = Object.keys(approvedLanguages)[0];
+    const newLanguage = Object.keys(newApprovedLanguages)[0];
     setIsTranslator(nextRole === "Translator");
     setLanguage(newLanguage);
-    setLevel(approvedLanguages[newLanguage]);
+
+    handleLevelDecrease(newApprovedLanguages[newLanguage]);
   };
 
   const handleLevelChangeStr = (nextLevel: string) => {
     setLevel(parseInt(nextLevel.replace("Level ", ""), 10));
   };
-
-  const approvedLanguages = role
-    ? approvedLanguagesTranslation
-    : approvedLanguagesReview;
 
   const languageOptions = isDisabled ? (
     <option key="undefined" value="undefined">
--- a/frontend/src/components/homepage/Filter.tsx
+++ b/frontend/src/components/homepage/Filter.tsx
@@ -13,6 +13,7 @@ import convertLanguageTitleCase from "../../utils/LanguageUtils";
 
 export type FilterProps = {
   approvedLanguagesTranslation: { [name: string]: number };
+  approvedLanguagesReview: { [name: string]: number };
   level: number;
   setLevel: (newState: number) => void;
   language: string;
@@ -24,6 +25,7 @@ export type FilterProps = {
 
 const Filter = ({
   approvedLanguagesTranslation,
+  approvedLanguagesReview,
   level,
   setLevel,
   language,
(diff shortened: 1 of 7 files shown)


frontend/src/components/homepage/Filter.tsx
```diff
@@ -32,39 +34,65 @@ const Filter = ({
...
-  const maxLvl = approvedLanguagesTranslation[language];
-  const lvlOptions = [];
-  for (let i = 1; i <= maxLvl; i += 1) {
-    lvlOptions.push(
-      <option key={i.toString()} value={i.toString()} label={i.toString()} />,
-    );
-  }
+
+  const maxApprovedLevel = approvedLanguages[language];
+  const levelOptions = Array.from(Array(maxApprovedLevel).keys()).map(
```
gaoxk, line 72: nit: IMO something like this is more clear for a one-liner array generator
```
const arr = [...Array(N+1).keys()].slice(1);
```
policy 71'd it from here https://www.techiedelight.com/create-array-from-1-n-javascript/
What do you think? Lots of interesting options on that site

gaoxk (changes requested):
noiiiice

frontend/src/components/homepage/Filter.tsx
Puepis: oh yeah I tried using the spread operator and the TS compiler was complaining so I went with `Array.from` instead

i'll try dealing with the ts error to use the spread operator bc I think it looks cleaner too

frontend/src/components/pages/HomePage.tsx
```diff
@@ -19,14 +19,16 @@ const HomePage = () => {
     authenticatedUser!!.approvedLanguagesTranslation.replace(/'/g, '"'),
   );
 
+  const approvedLanguagesReview = authenticatedUser!!.approvedLanguagesReview
```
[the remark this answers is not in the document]
Puepis: it's to handle the case of `approvedLanguagesReview` being null, initially I tried doing something like 
`const approvedLanguagesReview = JSON.parse(authenticatedUser!!.approvedLanguagesReview.replace(/'/g, '"')) ?? {}` but it causes a TypeError because of the `.replace` call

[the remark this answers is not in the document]
gaoxk: Ah ty for clarifying! make sense that there's a typescript error on the `.replace` call. Happy with the current code

frontend/src/components/homepage/Filter.tsx
gaoxk: linters do be raining on our parade 😢 feel free to strike a balance between making the linter happy and finding readable code. The current method is not horrendous by any means.

frontend/src/components/homepage/Filter.tsx
```diff
@@ -32,39 +34,65 @@ const Filter = ({
...
   const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     setLanguage(event.target.value);
   };
+
   const handleRoleChange = (nextRole: string) => {
+    const newApprovedLanguages =
+      nextRole === "Translator"
+        ? approvedLanguagesTranslation
+        : approvedLanguagesReview;
+    const newLanguage = Object.keys(newApprovedLanguages)[0];
```
jennchenn, line 50: When switching between translator<>reviewer, do we want to do a check first to see if the user has permission to translate/review in the same language that's currently selected instead of defaulting to the first language in the array? At least based on the current seed data, all the languages a user is able to review, they are able to translate as well--I was a bit confused when testing and switching around from reviewer to translator to see the language switch, but also not sure if that'd just overcomplicate with limited user experience value add

jennchenn (commented):
Just tested it out and it looks great 🥳 

I think it would be nice if we can show all the levels on the "My Work" tab, instead of just the levels the user is approved for in the language they had selected before switching to the "My Work" tab (i.e. when the filter `isDisabled=True` show levels 1-4)
image

frontend/src/components/homepage/Filter.tsx
gaoxk: good point - I think what jenn is describing is more user friendly

Puepis: yup I agree! i'll add that in

frontend/src/components/homepage/Filter.tsx
```diff
@@ -38,28 +38,25 @@ const Filter = ({
...
   const handleRoleChange = (nextRole: string) => {
     const newApprovedLanguages =
       nextRole === "Translator"
         ? approvedLanguagesTranslation
         : approvedLanguagesReview;
-    const newLanguage = Object.keys(newApprovedLanguages)[0];
+
+    let newLanguage = language;
+
+    // if user isn't permitted to translate/review in the same language, default
```
gaoxk, line 53: ty for documenting :D

frontend/src/components/homepage/Filter.tsx
```diff
@@ -79,9 +76,11 @@ const Filter = ({
   );
 
   const maxApprovedLevel = approvedLanguages[language];
-  const levelOptions = Array.from(Array(maxApprovedLevel).keys()).map(
-    (val) => `Level ${val + 1}`,
-  );
+  const levelOptions = isDisabled
+    ? ["Level 1", "Level 2", "Level 3", "Level 4"]
+    : [...Array(maxApprovedLevel + 1).keys()]
```
gaoxk, line 81: nit: add quick comment here that the else of the ternary is building a `Level 1 ... Level max` array

frontend/src/components/homepage/Filter.tsx
```diff
@@ -32,39 +34,74 @@ const Filter = ({
...
 }: FilterProps) => {
+  const approvedLanguages = role
+    ? approvedLanguagesTranslation
+    : approvedLanguagesReview;
+
   const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     setLanguage(event.target.value);
   };
+
   const handleRoleChange = (nextRole: string) => {
```
gaoxk, line 45: `setLevel(1)` should be used in this function,  on `planetread+miroslavklose@uwblueprint.org`, switching from translator to reviewer while on polish and lvl4 keeps the states on polish and lvl4.  `planetread+miroslavklose@uwblueprint.org` is only proved up to polish lvl 3 though

gaoxk: I think we also talked about using `setLevel(1)` on language change too on slack - correct me if im interpreting the current ux wrong

gaoxk (changes requested):
😎 lookin guuud

frontend/src/components/homepage/Filter.tsx
Puepis: oh yeah mb I think I interpreted what you said in slack wrong, I'll add it in
the UI issue with the radio group showing the wrong selected level still persists though - i'll include a recording

gaoxk (approved):
:shipit: :shipit: :shipit: :shipit: :shipit:

PR #104, opened 2021-09-25
Disable editing on translation platform if stage is REVIEW

lynn-kim:
## Notion ticket link

[Disable editing on translation platform if stage is REVIEW](https://www.notion.so/uwblueprintexecs/Disable-editing-on-translation-platform-if-stage-is-REVIEW-2b719a09478146e29d2c43d2cfa64c43)

## Implementation description

- disabled commenting, editing, and submitting when story translation is NOT in TRANSLATE state (instead of when story is in review)
- added tooltips 

## Steps to test

1. Set a story to be in REVIEW through mysql CLI 
2. "Send for review" button should now show "In review", table should not be editable, commenting should be disabled 
3. Tooltips should show up on these three areas 
4. Tooltips should NOT show up when in translate, buttons, commenting, and editing should work properly 

- I personally found that the tooltips on the text cells (in the table) were felt a bit unnecessary (and even distracting) --> would like to hear other people's thoughts 
- The tooltip label is currently repeated across three files but this was in case the copy needed to be different for different components 

## Checklist

### DIFF
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -14,7 +14,7 @@ import convertLanguageTitleCase from "../../utils/LanguageUtils";
 import deepCopy from "../../utils/DeepCopyUtils";
 import Header from "../navigation/Header";
 import CommentsPanel from "../review/CommentsPanel";
-import TRANSLATION_PAGE_TOOL_TIP_COPY from "../../utils/Copy";
+import { TRANSLATION_PAGE_TOOL_TIP_COPY } from "../../utils/Copy";
 
 type TranslationPageProps = {
   storyIdParam: string | undefined;
--- a/frontend/src/components/pages/TranslationPage.tsx
+++ b/frontend/src/components/pages/TranslationPage.tsx
@@ -1,7 +1,8 @@
 import React, { useState } from "react";
 import { useQuery } from "@apollo/client";
-import { Box, Button, Divider, Flex, Text } from "@chakra-ui/react";
+import { Box, Button, Divider, Flex, Text, Tooltip } from "@chakra-ui/react";
 import { useParams } from "react-router-dom";
+
 import ProgressBar from "../utils/ProgressBar";
 import TranslationTable from "../translation/TranslationTable";
 import UndoRedo from "../translation/UndoRedo";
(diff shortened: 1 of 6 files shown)


frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -59,6 +60,8 @@ const TranslationPage = () => {
   // Font Size Slider
   const [fontSize, setFontSize] = useState<string>("12px");
 
+  const [stage, setStage] = useState<string>("");
```
gaoxk, line 63: nit: move these lines to 43, just before the "AutoSave" comment and after the "Story Data" comment

frontend/src/components/translation/TranslationTable.tsx
```diff
@@ -60,10 +62,12 @@ const TranslationTable = ({
             fontSize={fontSize}
           />
         ) : (
-          <Text variant="cell" fontSize={fontSize}>
-            {" "}
-            {storyLine.translatedContent!!}{" "}
-          </Text>
+          <Tooltip hasArrow label={tooltipCopy}>
```
gaoxk, line 65: hmmmm now the reviewer also sees this tooltip, which is unnecessary. Can you modify the TranslationTable to be able to 
- have disabled cells and no tool tip
- have disabled cells and tool tip

gaoxk: eg http://localhost:3000/review/6/13 shows tool tips

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -119,12 +122,15 @@ const TranslationPage = () => {
   const clearUnsavedChangesMap = () => {
     setChangedStoryLines(new Map());
   };
+  const tooltipCopy =
```
gaoxk, line 125: Since we reuse `toolTipCopy` - we should store it in a `utils/Copy.ts` file as a const (something like `TRANSLATION_PAGE_TOOL_TIP_COPY`) and import it across these files.

gaoxk (changes requested):
noiiice! looks pretty good - happy to approve after iteration

frontend/src/utils/Copy.ts
```diff
@@ -0,0 +1,4 @@
+const TRANSLATION_PAGE_TOOL_TIP_COPY =
```
gaoxk, line 1: nit: change this to `export const TRANSLATION_PAGE_TOOL_TIP_COPY ....` rather than `export default` since this file will likely grow to export alot more copy.
change the related import lines to `import { TRANSLATION_PAGE_TOOL_TIP_COPY } from "../../utils/Copy";`

gaoxk (approved):
feel free to merge after addressing smol nit!

PR #116, opened 2021-10-08
Run pytest in circle ci

Puepis:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description

-

## Steps to test

1.

-

## Checklist


.circleci/config.yml
```diff
@@ -12,11 +12,34 @@ jobs:
...
+            docker-compose -f docker-compose.testing.yml up --abort-on-container-exit
+
+  build:
+    docker:
+      - image: cimg/base:2021.05
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+          docker_layer_caching: false
```
jennchenn, line 31: curious, why did you set `docker_layer_caching: false`?

according to the documentation [here](https://circleci.com/docs/2.0/docker-layer-caching/#remote-docker-environment) it seems like it might be set to false by default

jennchenn (commented):
🎉🎉 wooo!! a few nits + questions

didn't test out making a draft PR but i did try running the testing compose file locally (i had to switch the line endings to get it to work) + regular docker compose worked fine for me

.circleci/config.yml
Puepis: oh the original `build` step had this so it was copied over, since it's false by default I think we don't need to explicitly declare it

I'm not too sure about the benefit of turning DLC on but based on their documentation it says:
> When none of the layers in the image change between job runs, DLC pulls the layers from cache from the image that was built previously and reuses those instead of rebuilding the entire image.

so it might improve the speed of our builds. what do you think?

.circleci/config.yml
```diff
@@ -1,22 +1,43 @@
...
+  build:
+    docker:
+      - image: cimg/base:2021.05
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Create env file
+          command: touch .env
```
gaoxk, line 32: Is this necessary :0 Is circle ci writing environments into `.env`?

docker-compose.testing.yml
```diff
@@ -0,0 +1,31 @@
+version: '3.7'
+
+services:
+  py-backend:
+    command: /app/tools/wait-for-it.sh test-db:3306 -- pytest
```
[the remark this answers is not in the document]
Puepis: yup that's right!

.circleci/config.yml
jennchenn: update: our circleci plan doesn't support DLC :(

jennchenn (approved):
🚢

PR #119, opened 2021-10-09
Authenticate users for review/translation platforms

jennchenn:
## Notion ticket link

[Authenticate users for review/translation platforms](https://www.notion.so/uwblueprintexecs/Authenticate-users-for-review-translation-platforms-58cbbacb73374ba7b84c2a7231cb75b1)

## Implementation description

- Get `translatorId` and `reviewerId` in query when translation/review page is loaded, and check if it matches the id of the currently logged in user
- Show `NotFound` page if id does not match

## Steps to test

1. Log in as a user with both translation/review stories (e.g. Dwight D Eisenhower)
2. Click on view translation on of their translation stories and verify that the translation page loads
3. Enter the same URL manually and verify that it loads
4. Switch the URL from `/translation/` to `/review/` and verify that the page doesn't load
5. Try selecting a new story to translate and verify that the translation page loads
6. Repeat steps 2-5 but for the review platform

- `TranslationPage.tsx` and `ReviewPage.tsx`  
- Before the query is completed, only an empty div should be returned (this was done to prevent the 404 from showing right away)--is there something else that should be rendered instead?

## Checklist


frontend/src/components/pages/ReviewPage.tsx
```diff
@@ -85,81 +93,101 @@ const ReviewPage = () => {
...
-    >
-      <Header title={title} />
-      <Divider />
-      <Flex justify="space-between" flex={1} minHeight={0}>
-        <Flex width="100%" direction="column">
-          <Flex justify="space-between" alignItems="center" margin="10px 30px">
-            <FontSizeSlider setFontSize={handleFontSizeChange} />
-          </Flex>
-          <Divider />
+  const ReviewPageContent = () => {
```
gaoxk, line 96: nit: i think you could write this with no `return` keyword

```
const ReviewPageContent = () =>
 ( <div....
```

frontend/src/components/pages/ReviewPage.tsx
```diff
@@ -85,81 +93,101 @@ const ReviewPage = () => {
...
+    return (
+      <div>
```
gaoxk, line 98: nit: you could use a[ react fragment](https://reactjs.org/docs/fragments.html#short-syntax) here using `<></>`

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -217,106 +225,129 @@ const TranslationPage = () => {
...
-        />
-      )}
-    </Flex>
-  );
+        )}
+      </div>
+    );
+  };
+
+  return <div>{isLoading ? <div /> : <TranslationPageContent />}</div>;
```
gaoxk, line 350: might be able to get away with react fragments here too

frontend/src/components/pages/TranslationPage.tsx
```diff
@@ -217,106 +225,129 @@ const TranslationPage = () => {
...
-            >
-              <Box>
-                <Button
-                  colorScheme="blue"
-                  size="secondary"
-                  margin="0 10px 0"
-                  disabled={!editable}
-                  onClick={onSendForReviewClick}
+
+  const TranslationPageContent = () => {
```
gaoxk, line 229: ditto nit

frontend/src/components/pages/ReviewPage.tsx
```diff
@@ -85,81 +93,101 @@ const ReviewPage = () => {
...
-        />
-      </Flex>
-    </Flex>
-  );
+        )}
+      </div>
+    );
+  };
+
+  return <div>{isLoading ? <div /> : <ReviewPageContent />}</div>;
```
gaoxk, line 190: might be able to get away with react fragments here too

gaoxk (changes requested):
noiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiice down to merge after some nits 

> Before the query is completed, only an empty div should be returned (this was done to prevent the 404 from showing right away)--is there something else that should be rendered instead?

plain div is good for now! nothing's been designed out for it yet

gaoxk (approved):
noicee! 🚢

PR #123, opened 2021-10-11
Fix ButtonRadioGroup bug

lynn-kim:
## Notion ticket link

[Set default value in  ButtonRadioGroup](https://www.notion.so/uwblueprintexecs/Set-default-value-in-ButtonRadioGroup-271b5ab56e2346adb6c598d96efec24a)

## Implementation description

- Added an event listener for when the language is changed -- checks that if the selected level is not part of the options, it updates to the max available option 

## Steps to test
Oct-10-2021 18-57-45

1. Login as a user who has at least two different approved languages for either translation/review 
2. On the homepage, filter select the max level on the language with the higher level, and then switch to the language with a lower approved level

The level should automatically update to the highest possible level, and this should be accurately displayed in the displayed stories (look at badges). 

-

## Checklist

### DIFF
--- a/frontend/src/components/utils/ButtonRadioGroup.tsx
+++ b/frontend/src/components/utils/ButtonRadioGroup.tsx
@@ -1,6 +1,6 @@
 /*  eslint react/jsx-props-no-spreading: 0 */
 /*  eslint react/destructuring-assignment: 0 */
-import React from "react";
+import React, { useEffect } from "react";
 import {
   Box,
   Flex,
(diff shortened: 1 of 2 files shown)


frontend/src/components/utils/ButtonRadioGroup.tsx
```diff
@@ -47,6 +47,7 @@ export type ButtonRadioGroupProps = {
   isDisabled?: boolean;
   options: string[];
   onChange: (newState: string) => void;
+  language?: string;
```
gaoxk, line 50: Since we could use ButtonRadioGroup for other things in the future, mind finding a more agnostic name for the `language` prop? Thoughts on `value` or `dependentValue`?

gaoxk (changes requested):
nice! v elegant ✨

gaoxk (approved):
niiice

PR #132, opened 2021-10-16
Use correct badge styles

gaoxk:
image
image

## Notion ticket link

[Display correct language badge colors on UserTable](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=ba22ac93d09040ff8219c60ac7caabdb&p=92be83d79e2d4a49900ca36c358c0dde)
(did not make separate ticket for StoryCard)

## Implementation description

- use variants `language` and `stage`
- make and use variant `role`
- use `getLevelVariant`

## Steps to test

1. login
2. look at story cards. noice
3. go to /admin. look at manage translators and manage reviewers. noice

- does it look pwetty uwu

## Checklist

### DIFF
--- a/frontend/src/theme/components/Badge.ts
+++ b/frontend/src/theme/components/Badge.ts
@@ -23,6 +23,9 @@ const Badge = {
     stage: {
       backgroundColor: "blue.50",
     },
+    role: {
+      backgroundColor: "blue.50",
+    },
     language: {
       backgroundColor: "gray.200",
     },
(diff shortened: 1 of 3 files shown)


frontend/src/theme/components/Badge.ts
```diff
@@ -23,6 +23,9 @@ const Badge = {
     stage: {
       backgroundColor: "blue.50",
     },
+    role: {
```
gaoxk, line 26: this seems redundant since stage is the same color, but as we scale we could want to use separate variants and possibly want to change their colors.

jennchenn: stage/role do seem to serve the same purpose at the moment (i.e. is it translation/review) so i think keeping it the same color makes sense, but having a separate variant is reasonable (and also doesn't cause any issues anyway)
if we were to differentiate, i think it would be nice to have different colors for translator vs. reviewer instead, so users can more easily tell if (for example when viewing story cards) that is a story they are translating or reviewing

gaoxk: good idea! will keep in mind

PR #134, opened 2021-10-16
Return translator name and reviewer name in getStoryTranslations

jennchenn:
## Notion ticket link

[Return translator name and reviewer name in getStoryTranslations](https://www.notion.so/uwblueprintexecs/Return-translator-name-and-reviewer-name-in-getStoryTranslations-70e77a15524f4f8cbf11c1991450e8d7)

## Implementation description

- Updated `getStoryTranslations` to return translator/reviewer full name

## Steps to test

1. Login as admin
2. Test that translator and reviewer full names are returned in getStoryTranslations [query](http://localhost:5000/graphql?query=%7B%0A%20%20storyTranslations%20(storyTitle%3A%20%22and%22)%20%7B%0A%20%20%20%20title%0A%20%20%20%20id%0A%20%20%20%20description%0A%20%20%20%20level%0A%20%20%20%20language%0A%20%20%20%20stage%0A%20%20%20%20translatorId%0A%20%20%20%20reviewerId%0A%20%20%20%20translator%0A%20%20%20%20reviewer%0A%20%20%7D%0A%7D%0A)
3. Play around with the filters to ensure joining with users table didn't break anything

- is it unclear that `translator` and `reviewer` are names (would something like `translator_name` be better)?
   -  or should first name/last name be returned separately? (`translator_fname`, `translator_lname`)
- `StoryTranslationResponseDTO` is used in multiple queries; I did not modify anything other than `get_story_translations` so `translator`/`reviewer` will always be null in the response even if there is a translator/reviewer, is that okay? (e.g. in `storyTranslationsAvailableForReview`)

## Checklist

### DIFF
--- a/backend/python/app/graphql/types/story_type.py
+++ b/backend/python/app/graphql/types/story_type.py
@@ -75,6 +75,8 @@ class StoryTranslationResponseDTO(graphene.ObjectType):
     level = graphene.Int(required=True)
     num_translated_lines = graphene.Int()
     num_approved_lines = graphene.Int()
+    translator_name = graphene.String()
+    reviewer_name = graphene.String()
 
 
 class StoryTranslationUpdateStatusResponseDTO(graphene.ObjectType):
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -1,4 +1,5 @@
 from flask import current_app
+from sqlalchemy.orm import aliased
 
 from ...graphql.types.story_type import (
     StageEnum,
@@ -12,6 +13,7 @@ from ...models.story_content import StoryContent
 from ...models.story_translation import StoryTranslation
 from ...models.story_translation_content import StoryTranslationContent
 from ...models.story_translation_content_status import StoryTranslationContentStatus
+from ...models.user import User
 from ..interfaces.story_service import IStoryService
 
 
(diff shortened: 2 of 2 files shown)


backend/python/app/graphql/types/story_type.py
```diff
@@ -75,6 +75,8 @@ class StoryTranslationResponseDTO(graphene.ObjectType):
     level = graphene.Int(required=True)
     num_translated_lines = graphene.Int()
     num_approved_lines = graphene.Int()
+    translator = graphene.String()
+    reviewer = graphene.String()
```
gaoxk, line 79: > is it unclear that translator and reviewer are names (would something like translator_name be better)?

I'd prefer `translator_name` !

backend/python/app/services/implementations/story_service.py
```diff
@@ -152,23 +154,49 @@ def get_story_translations(
...
                 .order_by(Story.id)
                 .all()
             )
 
-            for i in range(len(story_translations)):
+            story_translations = []
+
+            for i in range(len(story_translation_data)):
+                story_translation, translator, reviewer = story_translation_data[i]
                 story = next(
```
gaoxk, line 180: using `next` is pretty massive brained :-0 

I'm not very familiar with next but it seems like it needs to be used with a generator for max efficiency, does the `stories` satisfy that? https://stackoverflow.com/questions/31040017/is-next-in-python-really-that-fast
Either way, still happy to keep using next :)

jennchenn: oh hmm since we are always expecting just one result from the filtering, i used `next()` instead of converting first to list and returning the item at 0 index:
`story = list(filter(lambda x: x.id == story_translation.story_id, stories))[0]` <- do you think this would be clearer?

inspiration from:
https://stackoverflow.com/questions/34026045/return-the-first-element-or-none-from-filter
https://docs.python.org/3/library/functions.html#next

gaoxk: `story = list(filter(lambda x: x.id == story_translation.story_id, stories))[0]` isn't particularly more clear imo

Happy with the code as is!

PR #136, opened 2021-10-17
Modify users query for admins

lynn-kim:
## Notion ticket link

[Modify users query for admins](https://www.notion.so/uwblueprintexecs/Modify-users-query-for-admins-55da95f592874338bb7180bdda4be0b0)

## Implementation description

- updated the users query and resolver to take more params

## Steps to test

1. Open apollo GUI or graphiql
2. Copy query [here](http://localhost:5000/graphql?variables=null&operationName=Users&query=query%20Users%20%7B%0A%20%20users(isTranslators%3Atrue)%20%7B%0A%20%20%20%20id%0A%20%20%20%20firstName%0A%20%20%20%20lastName%0A%20%20%20%20approvedLanguagesTranslation%0A%20%20%20%20approvedLanguagesReview%0A%20%20%7D%0A%7D)
3. see that the returned values are valid
4. try passing other variables (language, level, name/email) 
5. see that the returned values are valid

- querying works correctly
- would like feedback if there is a better way to pass params for buildUsersQuery (there's something similar in buildCommentsQuery but I thought this would be better since with three + params the if/else statements could get really long) 

## Checklist

### DIFF
--- a/frontend/src/APIClients/queries/UserQueries.ts
+++ b/frontend/src/APIClients/queries/UserQueries.ts
@@ -27,11 +27,9 @@ export const buildUsersQuery = (
   level?: number,
   nameOrEmail?: string,
 ) => {
-  const queryParams = `
-    ${language ? `language: "${language}", ` : ""} ${
-    level ? `level: ${level}, ` : ""
-  } ${nameOrEmail ? `nameOrEmail: "${nameOrEmail}", ` : ""}
-  `;
+  let queryParams = language ? `language: "${language}", ` : "";
+  queryParams += level ? `level: ${level}, ` : "";
+  queryParams += nameOrEmail ? `nameOrEmail: "${nameOrEmail}", ` : "";
 
   const queryString = gql`
     query Users {
--- a/backend/python/app/services/implementations/user_service.py
+++ b/backend/python/app/services/implementations/user_service.py
@@ -124,7 +124,7 @@ class UserService(IUserService):
         if language is not None and language is not "":
             query = query.filter(appr_langs.contains(language))
             if level:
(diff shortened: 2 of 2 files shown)


frontend/src/APIClients/queries/UserQueries.ts
```diff
@@ -9,24 +21,28 @@ export type User = {
...
 
-export const buildUsersQuery = () => {
-  // TODO: Modify when users query is modified
+export const buildUsersQuery = (
+  isTranslators: boolean,
+  language?: string,
+  level?: number,
+  nameOrEmail?: string,
+) => {
+  const queryParams = `
```
gaoxk, line 30: oof, the linter made quite a mess out of this. Maybe something like this could be more readable? what do you think?
```
let queryParams =  `${language ? `language: "${language}", ` : ""}`  
queryParams += `${level ? `level: ${level}, ` : ""}`
queryParams += `${nameOrEmail ? `nameOrEmail: "${nameOrEmail}", ` : ""}`;
```
alternatively, we can disable the linting in this code block and just straighten out the spacing a little:
```
/* eslint-disable my-error */
 const queryParams = `
 ${language ? `language: "${language}", ` : ""} 
 ${level ? `level: ${level}, ` : ""} 
 ${nameOrEmail ? `nameOrEmail: "${nameOrEmail}", ` : ""}
  `;
/* eslint-disable my-error */
```

gaoxk (changes requested):
nice! just gunna check in with designers on what the expected user behaviour is if only the level is selected

gaoxk (approved):
nice! lets merge it for now - we can update this later based on what the designers say about the level only filter

PR #147, opened 2021-11-01
return with level param only

andrewyguo:
## Notion ticket link

[:3](https://www.notion.so/uwblueprintexecs/Return-getUsers-with-level-param-only-aee4ad72b23947c182b9a6f951d1bcea)

## Implementation description

- in get_users() in `\backend\python\app\services\implementations\user_service.py`, filter the list of `user_dtos` by level before returning 

## Steps to test

1. Go to admin page and Manage Reviewers / Translators 
2. Only select the level filter parameter 
3. The resulting list of users should be all users that are approved for a language in that level 

- Small PR, just make sure nothing breaks 

## Checklist

[hi](https://www.economist.com/leaders/2020/10/29/why-it-has-to-be-biden)

### DIFF
--- a/backend/python/app/services/implementations/user_service.py
+++ b/backend/python/app/services/implementations/user_service.py
@@ -122,7 +122,7 @@ class UserService(IUserService):
 
         query = User.query.filter_by(role="User").filter(appr_langs.isnot(None))
 
-        if language is not None and language is not "":
+        if language:
             query = query.filter(appr_langs.contains(language))
             if level:
                 query = query.filter(appr_langs[language] == level)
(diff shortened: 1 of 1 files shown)


backend/python/app/services/implementations/user_service.py
```diff
@@ -150,6 +150,20 @@ def get_users(self, isTranslators, language=None, level=None, name_or_email=None
                 )
                 raise e
 
+        if (level is not None) and (not language or language == ""):
```
gaoxk, line 153: nit: `if level and not language` should be sufficient, `""` is falsey

gaoxk (approved):
feel free to merge after nit!

backend/python/app/services/implementations/user_service.py
andrewyguo: ah the conditional looks like that because I took the negation of this statement: 
image

backend/python/app/services/implementations/user_service.py
```diff
@@ -150,6 +150,20 @@ def get_users(self, isTranslators, language=None, level=None, name_or_email=None
                 )
                 raise e
 
+        if (level is not None) and (not language):
```
gaoxk, line 153: nit: `if level and (not language):` should be sufficient

gaoxk (approved):

PR #150, opened 2021-11-03
Soft delete users

lynn-kim:
## Notion ticket link

[Implement soft delete user mutation ](https://www.notion.so/uwblueprintexecs/Implement-soft-delete-user-mutation-6ec505fc7691411b9cea0a9c1d1afc5a)

## Implementation description

- Added database changes and migrations to create a users_all table and a users view table 
- Added soft delete mutation that also sets any story translation with the user translator/reviewer id to `null`, and soft deletes the comment that 

## Steps to test

**Test migration**
1. Rebuild docker containers and exec into backend container: `docker exec -it planet-read_py-backend_1 /bin/bash`
2. Clear db: `flask db downgrade base` / `python -m tools.insert_test_data erase`
3. Upgrade db: `flask db upgrade`
4. Seed db: `python -m tools.insert_test_data`

**Test delete**
1. Get access token by logging in as an admin and update headers in Altair
2. Soft delete a story translation:
3. 
```
mutation {
  softDeleteUser(id:1) {
    ok
  }
}
```

- check that the user doesn't show up in query `select * from users where id = 1;`, and that it does show up with is_deleted=1  (`select is_deleted from users_all where id =1;`)
- check that story_translations where the user was assigned is now `NULL` (from seed this should be rows 10 and 13)
- check that the comments that belonged to the user are soft deleted (`select is_deleted from comments_all where user_id = 1;`);

- Does the migration work?
- Does the soft delete for users work as expected, and update the correct values?

## Checklist


backend/python/app/services/implementations/user_service.py
```diff
@@ -275,6 +278,36 @@ def update_user_by_id(self, user_id, user):
 
         return UserDTO(user_id, user.first_name, user.last_name, user.email, user.role)
 
+    def soft_delete_user(self, user_id):
+        try:
+            user = User.query.get(user_id)
```
gaoxk, line 283: nit: I dont think users should be allowed to delete themselves, can you throw an exception if the user is trying to delete itself?

backend/python/app/services/interfaces/user_service.py
```diff
@@ -111,6 +111,17 @@ def update_user_by_id(self, user_id, user):
         """
         pass
 
+    @abstractmethod
+    def soft_delete_user(self, user_id):
+        """
+        Sft elete a user by user_id by setting is_deleted=True
```
gaoxk, line 117: nit: typo - Soft Delete a user ..

gaoxk (changes requested):

backend/python/app/services/implementations/user_service.py
lynn-kim: Fixed!

backend/python/tools/insert_test_data.py
```diff
@@ -174,7 +175,7 @@ def erase_db():
     db.session.commit()
     db.session.query(StoryTranslationAll).delete()
     db.session.commit()
-    db.session.query(User).delete()
```
gaoxk, line 177: Only weird behaviour i observed goes as follows
- `python -m tools.insert_test_data`
- soft delete a user using this pr's mutation
- `python -m tools.insert_test_data erase`
- soft deleted users still remain in the database
- `python -m tools.insert_test_data`
- no users are double inserted, but soft deleted users are still deleted (ie is_deleted=1)
any thoughts on this @jennchenn ? We dont have to address this issue in this pr but it is a little odd. I tried doing each of the following locally but the same behaviour still occurred:
```
def erase_db():
    ...
    db.session.query(User).delete()
    db.session.query(UserAll).delete()
    ...
```
and 
```
def erase_db():
   db.drop_all()
```

gaoxk (approved):
LGTM!!!

PR #158, opened 2021-11-14
Modify CircleCI to push code to client server pt1

jennchenn:
## Notion ticket link

[Modify CircleCI to push code to client server](https://www.notion.so/uwblueprintexecs/Modify-CircleCI-to-push-code-to-client-server-4a0f285dc95a483cb8970ca0e07c2b51)

## Implementation description

- modify circleci config to push backend code+frontend build files to the FTP server
- added appropriate FTP server credentials to circleci

## Steps to test

1. Verify that the release circleci deployment [succeeded](https://app.circleci.com/pipelines/github/uwblueprint/planet-read/1229/workflows/95bc206d-2598-4268-93fa-89a183cb3278/jobs/1820)
2. Check that the right files show up in the FTP server -> download the extension [here ](https://marketplace.visualstudio.com/items?itemName=Natizyskunk.sftp) and I'll send the config file!
image

- does it make sense to only deploy on `v.##` tags?
- what is a good name for the circleci workflow (if we continue adding steps, it'll turn kinda long `lint-build-test-deploy-....`)
- NOTE: once this is merged into main, will need to modify command to `git ftp push` instead of `init` (it must be initialized again in main so the commit hashes match up with main branch commits)
     - also just a personal note, before it is deployed from the main branch, the `.git-ftp.log` files must be deleted from the server so it can be re-initialized! 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,38 @@
 version: 2.1
 jobs:
+  lint:
+    docker:
+      - image: cimg/python:3.9.7
+    steps:
+      - checkout
+      - run:
+          name: Run linter checks
+          command: |
+            pip install black isort
+            echo "Running black checks..."
+            black backend/python/app/ --check
+            echo "Checking import order..."
+            isort --profile black backend/python/app/ -c
+            echo "Code properly linted!"
+
+  build:
+    docker:
+      - image: cimg/base:2021.05
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Create env file
+          command: touch .env # empty .env file needed to build containers
+      - run:
+          name: Build backend and frontend images
+          command: |
+            set -x
+            echo "Build images"
+            docker-compose build frontend 
+            docker-compose build py-backend
+
   test:
     docker:
       - image: cimg/base:2021.05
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -38,29 +33,93 @@ jobs:
...
+            sudo apt-get update
+            sudo apt-get -qq install git-ftp
+            echo "Setting up FTP credentials..."
+            HOST=`echo ${FTP_HOST_BASE64} | base64 --decode`
+            USERNAME=`echo ${FTP_USERNAME_BASE64} | base64 --decode`
+            PASSWORD=`echo ${FTP_PASSWORD_BASE64} | base64 --decode`
+            echo "export HOST=$HOST" >> $BASH_ENV
+            echo "export USERNAME=$USERNAME" >> $BASH_ENV
+            echo "export PASSWORD=$PASSWORD" >> $BASH_ENV
+      - run:
```
gaoxk, line 79: Interesting that you're deploying frontend and backend separately, any reason for this?

.circleci/config.yml
```diff
@@ -38,29 +33,93 @@ jobs:
...
+            git config git-ftp.password ${PASSWORD}
+            git ftp init --insecure --syncroot backend/python
+            echo "Backend deployed!"
 
 workflows:
   version: 2
   lint-build-test:
     jobs:
-      - lint
+      - lint:
```
gaoxk, line 102: why add the version tag filter to `lint` `build` and `test?`

frontend/src/components/review/ExistingComment.tsx
```diff
@@ -68,7 +68,12 @@ const ExistingComment = ({
         }
       }
     } catch (err) {
-      handleError(err ?? "Error occurred, please try again.");
```
gaoxk, line 71: is this left over correction from the old days of inconsistent linting :0

gaoxk (commented):
>does it make sense to only deploy on v.## tags?

yep that makes sense to me! We can to sprintwise-releases. mind adding some documentation to the README about setting up these releases? I think it would also make sense to come up with a versioning system. eg `v.1.x.y` where x is the sprint number and y is only greater than zero for immediate bug fixes. thoughts?

> what is a good name for the circleci workflow (if we continue adding steps, it'll turn kinda long lint-build-test-deploy-....)

`lint-build-test-deploy` is fine with me for now, anything beyond that and I think it's worth switching to something generic like "workflow". I dont think it would be too confusing to be too generic since there exists only one workflow. 


> NOTE: once this is merged into main, will need to modify command to git ftp push instead of init (it must be initialized again in main so the commit hashes match up with main branch commits)

Makes sense - I trust you'll have no problem making these changes before merging :) lmk if theres anything i can help with!

gaoxk:
also, that sftp extension is quite handy :^) great find!

.circleci/config.yml
jennchenn: I think the frontend build is supposed to be in the root of the server so the `index.html` can be served when a user goes to the website
Right now I just have the backend going to a folder called `backend/`; because they aren't being deployed in the same relative locations as our repository, I had to set them up separately

.circleci/config.yml
jennchenn: for some strange reason, because `lint`, `build` and `test` are all dependencies of `deploy`, the version tag filter doesn't work unless it's added to all of them
https://discuss.circleci.com/t/how-to-deploy-only-on-tagged-and-master-commits/25581/2

frontend/src/components/review/ExistingComment.tsx
jennchenn: i think so--i was getting an error when i was building locally that this wasn't linted properly, but the main firebase builds are passing hmm

README.md
```diff
@@ -121,6 +121,18 @@ The heroku app is the backend at the head of the main branch. The database is re
 
 Each pr has it's own firebase deployment. Each deployment points to the same heroku app at master, so frontend pr deployment links with any backend changes should be used with caution.
 
+# Production Deployment
```
gaoxk, line 124: niceee 😍

.circleci/config.yml
gaoxk: ohhh makes sense. I was confused about how filters-tags work. For anyone who reads this in the future, this is relevant: 
https://circleci.com/docs/2.0/configuration-reference/#tags

> CircleCI does not run workflows for tags unless you explicitly specify tag filters. Additionally, if a job requires any other jobs (directly or indirectly), you must specify tag filters for those jobs.

backend/python/app/models/user_all.py
```diff
@@ -8,7 +8,7 @@
 
 
 class UserAll(db.Model):
-    __tablename__ = "users"
+    __tablename__ = "users_all"
```
gaoxk, line 11: woah what 😳 im surprised this didnt cause any errors earlier. how did you discover this?

gaoxk (approved):
🚀 🚀 🚀 🚀 🚀  lets gooooooooooooooooooooo 🤩

backend/python/app/models/user_all.py
jennchenn: me too, found this from doing the story test architecture exercise and looking through the models
i wonder if this is what caused the pytest issues we had? (though i think switching over to using migrations was a good idea since it is a closer imitation of what we do in normal development anyway)

PR #177, opened 2021-11-25
Fix slider contrast

Puepis:
## Implementation description

- fix slider button contrast for approved languages table and font size slider

### before
image

### after
image
image

## Steps to test

1.

- 

## Checklist

### DIFF
--- a/frontend/src/components/admin/ApprovedLanguagesTable.tsx
+++ b/frontend/src/components/admin/ApprovedLanguagesTable.tsx
@@ -118,7 +118,7 @@ const ApprovedLanguagesTable = ({
             <SliderTrack height="6px">
               <SliderFilledTrack bg="blue.500" />
             </SliderTrack>
-            <SliderThumb />
+            <SliderThumb boxShadow="0px 0px 3px black" />
             {[1, 2, 3, 4].map((val) => (
               <SliderMark
                 key={val}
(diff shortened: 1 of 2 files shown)


gaoxk (commented):
thoughts on making a `SliderThumb.ts` file and setting the `boxShadow` in the `baseStyle?` that way whenever we make sliders the thumb is contrast-y by default. The only problem is that youll need to pick between 3px and 2px for the `boxShadow`

(see `Button.ts` for example)

Puepis:
> thoughts on making a `SliderThumb.ts` file and setting the `boxShadow` in the `baseStyle?` that way whenever we make sliders the thumb is contrast-y by default. The only problem is that youll need to pick between 3px and 2px for the `boxShadow`
> 
> (see `Button.ts` for example)

hmm I tried doing it by declaring this in `SliderThumb.ts` and importing it in `theme/index.ts`:
```
const SliderThumb = {
  baseStyle: {
    boxShadow: "0px 0px 3px black",
  },
  sizes: {
    small: {
      boxSize: 3,
      boxShadow: "0px 0px 2px black",
    },
  },
};
export default SliderThumb;
```
but the style isn't being applied for some reason - do I need to make some changes elsewhere?

gaoxk (approved):

PR #184, opened 2021-12-01
Add additional_experiences json column to Users

jennchenn:
## Notion ticket link

[Add additional_experiences json column to Users](https://www.notion.so/uwblueprintexecs/Add-additional_experiences-json-column-to-Users-9d37a6a629de45cfb0180437ff75706a)

## Implementation description

- add `additional_experiences` column to User model
- generated migration for User model renaming done in #158 

## Steps to test

1. Run migrations: `docker exec -it planet-read_py-backend_1 /bin/bash -c "flask db upgrade"`
2. Verify that everything works as expected, and `additional_experiences` column appears in `users` table: `docker exec -it planet-read_db_1 mysql -u root -proot -e "USE planet-read; SELECT * FROM users;"`
3. Test that you can still create a new user by signing up for the platform via the frontend
4. Test that you are able to update an existing user:
```
mutation {
  updateUser(id: 1, user: {firstName: "Carl", lastName: "Sargan", role: User, email: "planetread+carlsagan@uwblueprint.org", additionalExperiences: "{\"language_experience\": \"none\"}"}) {
    user {
      firstName
      lastName
      email
      additionalExperiences
      approvedLanguagesReview
      approvedLanguagesTranslation
    }
  }
}
```

- recreated the view in the migration to ensure the column is picked up when a user queries the `users` table
- for the update mutation, if the user does not pass in a parameter, it'll set it to be null in the database--is this the behavior we want/should we change it so it only updates parameters that are specified directly (to either a new value or to null)?
https://github.com/uwblueprint/planet-read/blob/3bca26b5f4c97f3a915f08e084cf82785841c741/backend/python/app/services/implementations/user_service.py#L243-L253

## Checklist


gaoxk (approved):
nice! just an fyi, signup/login is broken with google rn. This issue is happening on main branch so no need to investigate in this pr. 

> for the update mutation, if the user does not pass in a parameter, it'll set it to be null in the database--is this the behavior we want/should we change it so it only updates parameters that are specified directly (to either a new value or to null)?

good catch! let's see where app development takes us and if we'll need to be concerned with this

PR #186, opened 2021-12-05
Update User Profile

lynn-kim:
## Notion ticket link

[Add gql mutation to update user profile ](https://www.notion.so/uwblueprintexecs/Add-gql-mutation-to-update-user-profile-3bb87ef76ece4c1c94fa6550b238a14f)

## Implementation description

- added a new gql for users to update their own profiles 
   - Currently getting the user id in the graphql stage and passing, since importing the auth middleware into user_service creates a circular dependency -- would there be a better way to do this?
- require authentication for the update user by ID mutation (for admin)

## Steps to test

1. Go to altair / or some other graphql gui 
2. Login as a [User](http://localhost:5000/graphql?variables=null&operationName=undefined&query=%0Amutation%20%7B%0A%20%20login%20(email%3A%22planetread%2Bmiroslavklose%40uwblueprint.org%22%2C%20password%3A%22dolphins_are_cool%22)%20%7B%0A%20%20%20%20accessToken%0A%20%20%20%20refreshToken%0A%20%20%20%20id%0A%20%20%7D%0A%7D) and set the header 
3. Try different mutations ([example](http://localhost:5000/graphql?variables=null&operationName=updateMe&query=mutation%20updateMe%20%7B%0A%20%20updateMe(user%3A%7BfirstName%3A%22miroslav%22%2C%20lastName%3A%22klose%22%2C%20role%3AUser%2C%20email%3A%20%22planetread%2Bmiro%40uwblueprint.org%22%7D)%7B%0A%20%20%20%20user%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20firstName%0A%20%20%20%20%20%20lastName%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D)) by adding different fields -- users should not be able to update role and approved languages on the updateMe mutation 

-

## Checklist

### DIFF
--- a/backend/python/app/graphql/mutations/user_mutation.py
+++ b/backend/python/app/graphql/mutations/user_mutation.py
@@ -1,6 +1,7 @@
 import graphene
 
 from ...middlewares.auth import (
+    get_user_id_from_request,
     require_authorization_by_role_gql,
     require_authorization_by_user_id_not_equal,
 )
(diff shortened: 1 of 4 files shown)


gaoxk (changes requested):
tested things out and it looks good!
seems odd that `role` has to be set on the `user` param passed to `updateMe`. thoughts on modifying `UserDTO`'s `role` to be non required?'

also just a little fyi - when updating a user's email, it updates the email on the firebase side too. so other devs doing work might start seeing issues with test accounts if the emails are outta sync. I corrected angelamerkel just now as it was the only one that looked changed on firebase. Could you double check all the test account emails are currently correct? https://console.firebase.google.com/u/5/project/planet-read-uwbp/authentication/users

lynn-kim:
> also just a little fyi - when updating a user's email, it updates the email on the firebase side too. so other devs doing work might start seeing issues with test accounts if the emails are outta sync. I corrected angelamerkel just now as it was the only one that looked changed on firebase. Could you double check all the test account emails are currently correct? https://console.firebase.google.com/u/5/project/planet-read-uwbp/authentication/users

oops they should be fixed now!!

gaoxk (approved):
👍 👍 👍 👍 👍 👍 sickk

PR #187, opened 2021-12-06
Implement assign story flow

Puepis:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Implement-Assign-new-story-flow-d09abaf754ee453985c4db4aab180ad5)

## Implementation description

- implement assign story modal + flow
- added check to assign_user_as_reviewer to prevent translator and reviewer being the same user

## Steps to test

- To assign a user, go to localhost:3000/user/:id and click "Assign New Story"
- Feel free to test any of the flows below or any other combination!

While playing around with the modal, verify that:
1. all other fields are disabled before role is selected
2. story field is disabled if the other fields aren't selected
3. only approved languages/levels are displayed

#### Successful Reviewer Assign

1. Go to /user/4 and click "Assign New Story"
2. select "Reviewer", "English (UK)", "3", and "A Tale of Two Cities"
3. after confirming, verify that the page reloads and the new row shows up on the table

#### Successful Translator Assign

1. Go to /user/4 and click "Assign New Story"
2. select "Translator", "English (US)", "1" and "War and Peace"
3. after confirming, verify that the page reloads and the new row shows up on the table

#### Fail Reviewer Assign

1. Go to /user/4 and click "Assign New Story"
2. select "Reviewer", "English (UK)", "3" and "To Kill a Mockingbird"
3. verify that the operation fails (the user is already reviewing an English (UK) story)

#### Fail Translator Assign

1. Go to /user/4 and click "Assign New Story"
2. select "Translator", "English (UK)", "1" and "War and Peace"
3. verify that the operation fails (the user is already translating another English (UK) story)

- pretty bulky PR but the main file to check is `AssignStoryModal.tsx`
- does everything work and is the modal intuitive to use?

## Checklist

### DIFF
--- a/frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
+++ b/frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
@@ -1,4 +1,5 @@
 import React, { useState, Dispatch, SetStateAction } from "react";
+import { useMutation } from "@apollo/client";
 import { Icon } from "@chakra-ui/icon";
 import { MdDelete } from "react-icons/md";
 import {
@@ -14,10 +15,23 @@ import {
   IconButton,
 } from "@chakra-ui/react";
 import { StoryTranslation } from "../../APIClients/queries/StoryQueries";
+import {
+  ASSIGN_REVIEWER,
+  AssignReviewerResponse,
+  CREATE_TRANSLATION,
+  CreateTranslationResponse,
+} from "../../APIClients/mutations/StoryMutations";
 import { getLevelVariant } from "../../utils/StatusUtils";
 import convertStageTitleCase from "../../utils/StageUtils";
 import { convertLanguageTitleCase } from "../../utils/LanguageUtils";
-import { generateSortFn } from "../../utils/Utils";
+import { ApprovedLanguagesMap, generateSortFn } from "../../utils/Utils";
+import ConfirmationModal from "../utils/ConfirmationModal";
+import {
+  ASSIGN_STORY_BUTTON,
+  ASSIGN_STORY_CONFIRMATION,
+} from "../../utils/Copy";
+import AssignStoryModal, { StoryToAssign } from "../utils/AssignStoryModal";
+import { StoryAssignStage } from "../../constants/Enums";
 
 interface AssignedStoryTranslationsFieldSortDict {
   [field: string]: {
@@ -31,17 +45,47 @@ export type AssignedStoryTranslationsTableProps = {
   storyTranslations: StoryTranslation[];
   setStoryTranslations: (newStoryTranslations: StoryTranslation[]) => void;
   userId: number;
+  approvedLanguagesTranslation: ApprovedLanguagesMap | undefined;
+  approvedLanguagesReview: ApprovedLanguagesMap | undefined;
+  setStoryAssignStage: (newStage: StoryAssignStage) => void;
 };
 
 const AssignedStoryTranslationsTable = ({
   storyTranslations,
   setStoryTranslations,
   userId,
+  approvedLanguagesTranslation,
+  approvedLanguagesReview,
+  setStoryAssignStage,
 }: AssignedStoryTranslationsTableProps) => {
   const [isAscendingTitle, setIsAscendingTitle] = useState(true);
   const [isAscendingRole, setIsAscendingRole] = useState(true);
   const [isAscendingLanguage, setIsAscendingLanguage] = useState(true);
   const [isAscendingStage, setIsAscendingStage] = useState(true);
+
+  const [assignStory, setAssignStory] = useState(false);
+  const [confirmAssignStory, setConfirmAssignStory] = useState(false);
+
+  const [storyToAssign, setStoryToAssign] = useState<StoryToAssign | null>(
+    null,
+  );
+
+  const closeAssignStoryModal = (cancelled = true) => {
+    setAssignStory(false);
+    if (cancelled) {
+      setStoryAssignStage(StoryAssignStage.CANCELLED);
+    }
+  };
+  const openAssignStoryModal = () => {
+    setAssignStory(true);
+  };
+
+  const closeConfirmAssignStoryModal = () => {
+    setConfirmAssignStory(false);
+  };
+  const openConfirmAssignStoryModal = () => {
+    setConfirmAssignStory(true);
+  };
 
   const getRole = (translation: StoryTranslation) => {
     if (userId === translation.translatorId) {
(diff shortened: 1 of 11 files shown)


frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
```diff
@@ -91,6 +136,68 @@ const AssignedStoryTranslationsTable = ({
...
+        language: storyToAssign!!.language,
+      };
+      const result = await createTranslation({
+        variables: { storyTranslationData },
+      });
+      return result.data?.createStoryTranslation.story.id
+        ? null
+        : "Error assigning user as translator!";
+    } catch (err) {
+      return typeof err === "string"
```
gaoxk, line 174: this is an interesting pattern. Why not just `return err` ?

frontend/src/components/utils/AssignStoryModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+        <ModalHeader paddingTop="20px">
+          <Heading as="h3" size="lg">
+            Assign Story
+          </Heading>
+          <Text fontSize="16px" fontWeight="normal">
+            Please assign an existing story to the user.
+          </Text>
+        </ModalHeader>
+        <ModalBody paddingTop="24px" paddingBottom="36px">
+          <Grid
```
gaoxk, line 155: nice to see Grid in use in our code base!

frontend/src/components/utils/AssignStoryModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+  // display levels that the user is approved for
+  const levelOptions = language
+    ? [...Array(approvedLanguages[language] + 1).keys()]
+        .slice(1)
+        .map((value) => ({ value: `${value}` }))
+    : [];
+
+  const disabledStyle = useStyleConfig("Disabled");
+  const isPropertySelectDisabled = role == null;
+  const isTitleSelectDisabled = language == null || level == null;
```
gaoxk, line 130: nit: theres a gap of time between the language&level being non null and having the actual titles retrieved from our api. in this time gap, the title select is enabled despite a lack of titles. Can we make this check on the length of the `stories` in addition to language&level being selected?

gaoxk (changes requested):
Nice work! It's a challenging pr. 

flow thoughts:
- doesnt make sense to enable the level select when the language has not been selected. there is no level to select from anyways. i would prefer the dropdown be disabled
- if a user is only approved to for one role, the role should be 'locked in', ie the admin shouldnt be able to use the drop down. Feel free to leave a todo for this if you think in can be addressed in a separate short pr

Bug I've encountered
- http://localhost:3000/user/2 
- assign new story flow
- select "translator" and "german"
- switch "translator" to "reviewer" (user only approved for polish german)
- also happens for user 1
image

I believe an intuitive behavior in this situation would be to clear all the fields and make the admin select language and level again if the language & level are not also approved for reviewing

And just a thought for the future: I think it's pretty user unfriendly that we can allow admins to try to assign stories for role+languages that the user cant take more translations for. I still think all the error handling you implemented is necessary and good code practice, but admins shouldnt be put in this situation. I dont think an extra TODO comment is needed in this pr to make note of this -- I'll think about a solution and draft a ticket for it.

frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
Puepis: oh I got the pattern from [here](https://github.com/uwblueprint/planet-read/blob/main/frontend/src/components/homepage/StoryCard.tsx#L73), I can change it to just `return err` tho

frontend/src/components/utils/InfoAlert.tsx
```diff
@@ -0,0 +1,24 @@
...
+const InfoAlert = ({ message }: AlertProps) => {
+  return (
+    <Alert
+      status="info"
+      borderRadius={8}
+      bg="gray.200"
+      justifyContent="center"
+      textColor="#858585"
+    >
+      <AlertIcon as={MdInfoOutline} color="#858585" />
```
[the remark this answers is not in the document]
Puepis: currently it's not defined in `colors.ts`, should I just add it as 600 in the object? 
```
gray: {
    50: "#F9FAFC",
    100: "#F1F4F7",
    200: "#DCDCDC",
    300: "#CECECE",
    400: "#606062",
    500: "#1C1C1E",
  }
```

Puepis:
> * if a user is only approved to for one role, the role should be 'locked in', ie the admin shouldnt be able to use the drop down. Feel free to leave a todo for this if you think in can be addressed in a separate short pr

hm what should the UI look like for this? would it just be a text field without the dropdown arrow?

Puepis:
> I believe an intuitive behavior in this situation would be to clear all the fields and make the admin select language and level again if the language & level are not also approved for reviewing

in that case would it be better to clear all the fields every time the role is changed?

frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
gaoxk: i like how clean `return err as string` is

Puepis:
mb I pretty much have the fixes done, gonna check with the designers about the disabled dropdown first

frontend/src/components/utils/AssignStoryModal.tsx
Puepis: @gaoxk if we add the check for length then the title select will be disabled if there are no stories to show, would that be misleading to the user (as opposed to having a dropdown with no options)?

gaoxk: hmmm I see what you mean. I think this would be effective and consider your described case:
`const isTitleSelectDisabled = (language == null || level == null) && stories.length > 0;`

frontend/src/components/utils/InfoAlert.tsx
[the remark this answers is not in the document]
gaoxk: (resolved discussion on slack.)

frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable */
```
gaoxk, line 1: I think this turns off eslint for the whole file, whats the specific line you want eslint to ignore?

frontend/src/components/utils/AssignStoryModal.tsx
```diff
@@ -0,0 +1,281 @@
+/* eslint-disable */
```
gaoxk, line 1: ditto comment about eslint

gaoxk (changes requested):
the disabled role dropdown is working nicely!
I'm still experiencing the bug i originally described where switching the role causes the app to crash

frontend/src/components/admin/AssignedStoryTranslationsTable.tsx
Puepis: oops forgot to remove from before

gaoxk (approved):
🚀 niceee!!!!

PR #204, opened 2021-12-18
Add translator_last_activity and reviewer_last_activity time columns…

jennifer-lu:
… to StoryTranslation model

## Notion ticket link

[Add translator_last_activity and reviewer_last_activity time columns to StoryTranslation model](https://www.notion.so/uwblueprintexecs/Add-translator_last_activity-and-reviewer_last_activity-time-columns-to-StoryTranslation-model-35720c1545224b138eaf3d0c6cf35ade)

## Implementation description

- Added time columns to database
- Created new update activity mutation that is called after:
    - storytranslationcontent edits
    - storytranslationcontent status changes
    - comments made
    - comments resolved

## Steps to test

1. Check that the columns `reviewer_last_activity` and `translator_last_activity` have been added to the `story_translations` and `story_translations_all` tables using mysql
2. The `reviewer_last_activity` and `translator_last_activity` columns should have null entries
3. Login as `planetread+carlsagan@uwblueprint.org`
4. Edit Great Gatsby translation
5. View the `story_translations` table again, the appropriate timestamp (row: id 10, column: translator_last_activity) should be updated (note that the time zone is utc)
6. Make a comment on Great Gatsby translation
7. View the `story_translations` table again, the appropriate timestamp (row: id 10, column: translator_last_activity) should be updated (note that the time zone is utc)
8. Send Great Gatsby translation for review
9. Sign out
10. Login as `planetread+dwightdeisenhower@uwblueprint.org`
11. Update status on Great Gatsby translation
12. View the `story_translations` table again, the appropriate timestamp (row: id 10, column: reviewer_last_activity) should be updated (note that the time zone is utc)
13. Make a comment on Great Gatsby translation
14. View the `story_translations` table again, the appropriate timestamp (row: id 10, column: reviewer_last_activity) should be updated (note that the time zone is utc)
15. You can repeat these steps with other users and translations as well

- Does the time zone make sense? (I am using utc to align with the comment timestamps)
- Resolving comments seems to be WIP, so I didn't test it (please let me know if it should be working)
- Does the implementation make sense? I created a `UpdateStoryTranslationLastActivity` mutation which is called after every applicable mutation (`UPDATE_STORY_TRANSLATION_CONTENTS`, `UPDATE_STORY_TRANSLATION_CONTENT_STATUS`, `CREATE_COMMMENT`, `UPDATE_COMMENT_BY_ID`)

## Checklist


gaoxk (changes requested):
Nice work so far! I tested as per your instructions and it made sense to me.

> Does the time zone make sense? (I am using utc to align with the comment timestamps)

yep! looks good to me

> Resolving comments seems to be WIP, so I didn't test it (please let me know if it should be working)

Resolving shouldn't be WIP, please lmk what you see that seems unfinished. There's some odd behaviour happening  when I resolve comments (the invalid time component showed up after resolving a comment) 

image

> Does the implementation make sense? I created a UpdateStoryTranslationLastActivity mutation which is called after every applicable mutation (UPDATE_STORY_TRANSLATION_CONTENTS, UPDATE_STORY_TRANSLATION_CONTENT_STATUS, CREATE_COMMMENT, UPDATE_COMMENT_BY_ID)

The current implementation in this pr requires two separate network calls each time the user interacts with a story translation (the original call + the UpdateStoryTranslationLastActivity call), which slows down the app. This can be simplified to a single gql call. For example, see `GET_STORY_AND_TRANSLATION_CONTENTS`. It completes queries `storyById` and `storyTranslationById` in one call.

However, `UpdateStoryTranslationLastActivity` introduces alot of frontend overhead (eg requiring `storyTranslationId` and `isTranslator` props in AutoSave, WIPComment, StatusBadge, etc). Additionally, the mutation has no authentication (ie no `@require` tags for roles) so it introduces the possibility of a random person on the internet abusing the endopint. Finally, because this is a seperate mutation from the actual interactions with the story translation, there is a possibility of the translator/reviewer abusing the endpoint to appear to be updating the storytranslation when they actually are not. 

To avoid creating the mutation entirely, the `update_story_translation_last_activity` service function can be called inside the relevant service functions or GQL mutations/queries. The relevant service functions are
- update_story_translation_contents
- update_story_translation_stage
- update_story_translation_content_status
- approve_all_story_translation_content
- create_comment
- update_comment

The services are each a little different, so deducing the `story_translation_id` and whether the caller is a reviewer or translator is different depending on the service function. 
**Example: `update_story_translation_contents`**
For example, `update_story_translation_contents` is used in the `UpdateStoryTranslationContents` mutation which is authenticated by `@require_authorization_as_story_user_by_role(as_translator=True)`. Therefore the mutation is always called by a translator, so the 
**Example: `update_story_translation_stage`**
line 456 has `user_id = int(get_user_id_from_request())`, so the role of the calling user could be determined from this id.

Unfortunately this is all quite confusing because we are not consistent in where we authenticate users. `update_story_translation_contents` authenticates at the gql level, but  `update_story_translation_stage` authenticates at the story_service level. We should really move the authentication to the GQL side to make sure all the services can be tested with pytest. I will work on this ASAP.

Therefore, I think its fine if you revert all changes except the database migration for this pr, and then edit the services/mutations in another pr. I think it's getting quite complicated. 

lmk if you have any questions, happy to hop on a call!! sorry for all the confusion and redundant work

PR #242, opened 2021-12-24
Implement multi thread comment styling

gaoxk:
## Notion ticket link
https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=ba22ac93d09040ff8219c60ac7caabdb&p=802f3a1377064bd79778c2a4ec31b616

## Implementation description

- defined and mapped comments to be thread heads or non thread heads
- styled comments accordingly 
- bonus: solved odd resolving behaviour 

A thread head is the beginning of a thread. it is not indented.
A live thread head is an unresolved thread head. it has a reply and resolve button. 

## Steps to test

1. `docker exec -it planet-read_py-backend_1 /bin/bash -c "python -m tools.insert_test_data erase"`
2. `docker exec -it planet-read_py-backend_1 /bin/bash -c "python -m tools.insert_test_data"`
3. login as carlsagan and go to http://localhost:3000/review/6/13
4. set comment filter to all
5. observe resolved thread on line 1
6. make new comment on line 1
7. observe that it appears as a live thread head
8. reply to the comment
9. observe that the replies appear as replies
10. resolve the thread. observe how it flattens into the resolved thread immediately
11. make new comment on line 1. observe that it appears as a live thread head. feel free to repeat steps 5-9 on filters unresolved & all. 
12. play with filters to see that comments still display correctly

Heres what I observe when testing:
image

**bonus round: wackass resolving behaviour fixed**
1. set filter to All
2. resolve an existing comment. observe that the reply/resolve buttons disappear
3. create a new comment and resolve it. observe that the reply/resolve buttons disappear
4. set filter to unresolved
5. resolve an existing comment. observe that the comment disappears
6. create a new comment and resolve it. observe that the comment disappears

- does it work
- does the code make sense? the comment section code is quite busy

## Checklist

### DIFF
--- a/frontend/src/components/review/CommentsPanel.tsx
+++ b/frontend/src/components/review/CommentsPanel.tsx
@@ -34,6 +34,7 @@ const CommentsPanel = ({
 }: CommentPanelProps) => {
   const [comments, setComments] = useState<CommentResponse[]>([]);
   const [filterIndex, setFilterIndex] = useState(2);
+  const [threadHeadMap, setThreadHeadMap] = useState<boolean[]>([]);
 
   const filterOptions = ["All", "Resolved", "Unresolved"];
   const filterToResolved = [null, true, false];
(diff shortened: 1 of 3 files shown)


frontend/src/components/review/CommentsPanel.tsx
```diff
@@ -34,17 +34,43 @@ const CommentsPanel = ({
...
+        cmts.findIndex(
+          (_c) => _c.storyTranslationContentId === c.storyTranslationContentId,
+        )
+      )
+        return true;
+      const lastC = cmts[i - 1];
+      return (
+        !c?.resolved &&
+        lastC?.resolved &&
+        lastC?.storyTranslationContentId === c?.storyTranslationContentId
```
Puepis, line 63: just to confirm: this condition handles the case where the current comment thread corresponds to the same story line as the previous comment?

Puepis (changes requested):
when I test the flow I get some weird behaviour where only the comment gets resolved (and not the replies), but it corrects itself after switching filters/refreshing:
behaviour

frontend/src/components/review/CommentsPanel.tsx
gaoxk: yesssssssssssssss. if the previous comment is resolved but the current one is not, and they are both for the same content line, then the current one is the start of a new thread

gaoxk:
thanks for catching that case! Updated to handle resolving threads under the "all" filter

Puepis (approved):
played around with the comment resolving flow and it looks good to me! just had a couple of questions about the commenting behaviour:

1. if the user has the "Resolved" filter on and creates a new comment, the new comment will still appear in the panel even though it's unresolved - is this the expected behaviour? (personally I think it's okay because it shows the user that the comment has actually been created)
2. if a user comments on a line then repeats the same flow, the second comment gets treated and displayed as a reply to the first - is this expected?

PR #244, opened 2021-12-26
Download files via fileById

Puepis:
## Notion ticket link

https://www.notion.so/uwblueprintexecs/Download-files-via-fileById-cce5b4acf4b0455981e6452e294eec77

## Implementation description

- implement `fileById` query by converting file data to a base64 string and returning it in a `DownloadFileDTO`

## Steps to test

1. start docker containers
2. in any directory, run `curl http://localhost:5000/graphql -F operations='{"query": "mutation ($file: Upload!) { createFile(file: $file) { ok }}", "variables": {"file": null }}' -F map='{ "0": ["variables.file"]}' -F 0=@file_name.txt`, replacing `file_name.txt` with any file in the directory to upload the file
3. in Altair, login as any user and run the query below
4. verify that the query succeeds and the file string is returned
5. if you want to verify that the string encoded properly, copy and run the following code in `codesandbox.io`

```
// Retrieved from https://developer.mozilla.org/en-US/docs/Glossary/Base64
function b64_to_utf8(str) {
  return decodeURIComponent(escape(atob(str)));
}
const b64String = "FILE_STRING_HERE";
const decoded = b64_to_utf8(b64String);
console.log(decoded);
```
**Note**: I'm not sure if the above ^ works for decoding non-`.txt` files

#### fileById query
```
query Download {
  fileById(id: 1) {
    id
    file
  }
}
```

- does it work for files that aren't `.txt` files? e.g. `.pdf`, `.docx`

Also, should we include a `fileName` field in the `DownloadFileDTO`? I'm thinking it'll be useful to at least return the file extension

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/file_service.py
+++ b/backend/python/app/services/implementations/file_service.py
@@ -22,7 +22,10 @@ class FileService(IFileService):
             # encode file data as base64 string
             file_bytes = open(file.path, "rb").read()
             encoded = b64encode(file_bytes).decode("utf-8")
-            return DownloadFileDTO(id=id, file=encoded)
+
+            # extract extension and path
+            _, file_extension = os.path.splitext(file.path)
+            return DownloadFileDTO(id=id, file=encoded, ext=file_extension[1:])
         except Exception as e:
             reason = getattr(e, "message", None)
             self.logger.error(
--- a/backend/python/app/services/implementations/file_service.py
+++ b/backend/python/app/services/implementations/file_service.py
@@ -1,5 +1,7 @@
 import os
+from base64 import b64encode
 
+from ...graphql.types.file_type import DownloadFileDTO
 from ...models import db
 from ...models.file import File
 from ..interfaces.file_service import IFileService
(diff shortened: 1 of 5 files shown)


gaoxk (commented):
works as outlined!

Question: Imagine our frontend makes this gql request. With the decoded information, how would the frontend save the file? Would the frontend need to know what kind of file (eg txt vs pdf) it is receiving?

Puepis:
> With the decoded information, how would the frontend save the file? 

I did some searching and it seems like we would need to convert the base64 string to a blob, then either use a library like [file-saver](https://www.npmjs.com/package/file-saver) or use `URL.createObjectURL(new Blob([blob]))` to create a link and inject it into the DOM

Some links: 
- https://javascript.plainenglish.io/how-to-create-download-and-upload-files-in-react-apps-80893da4247a
- https://stackoverflow.com/a/63965930

> Would the frontend need to know what kind of file (eg txt vs pdf) it is receiving?

yup I believe so, I can add a field called `ext` to the DTO to return the file extension

backend/python/app/services/implementations/file_service.py
```diff
@@ -22,7 +22,10 @@ def get_file(self, id):
             # encode file data as base64 string
             file_bytes = open(file.path, "rb").read()
             encoded = b64encode(file_bytes).decode("utf-8")
-            return DownloadFileDTO(id=id, file=encoded)
+
+            # extract extension and path
```
gaoxk, line 26: nit: drop the `.` before the extension
```
{
  "data": {
    "fileById": {
      "id": 1,
      "ext": ".txt"
    }
  }
}
```

gaoxk (approved):
noice

PR #267, opened 2021-12-27
Enable admin to download resume

Puepis:
## Notion ticket link

https://www.notion.so/uwblueprintexecs/Enable-admin-to-download-resume-f1ddd118506b466b8e33b1b376ee8467

## Implementation description

- add "User files" sidebar section to `UserProfilePage.tsx` for admins
- implement utility functions to download files
- created `useFileDownload` hook to encapsulate generic file download logic

for backend implementation, see https://github.com/uwblueprint/planet-read/pull/244

## Steps to test

1. `docker-compose up --build`
2. Upload a few files (ideally pdf) using curl: `curl http://localhost:5000/graphql -F operations='{"query": "mutation ($file: Upload!) { createFile(file: $file) { ok }}", "variables": {"file": null }}' -F map='{ "0": ["variables.file"]}' -F 0=@file_name.pdf`
3. set resume of carl sagan to file id: `docker exec -it planet-read_db_1 mysql -u root -proot -e "UPDATE users SET resume=FILE_ID_HERE WHERE id=1;"`
4. Login as admin and go to `http://localhost:3000/#/user/1`
5. Click on "resume.pdf"
6. Verify that the file is downloaded and is the same as the file you uploaded
7. feel free to upload and test with different files (i tested with pdf, docx, txt)

- does it work?
- does the `useFileDownload` hook make sense?
- how should we display the filename under "User files"? right now it's hardcoded to resume.pdf` but another option is to fetch the complete filename from the backend instead of just the extension

image

## Checklist

### DIFF
--- a/frontend/src/utils/FileUtils.ts
+++ b/frontend/src/utils/FileUtils.ts
@@ -1,0 +1,72 @@
+import { useLazyQuery } from "@apollo/client";
+import { saveAs } from "file-saver";
+import { File, GET_FILE } from "../APIClients/queries/FileQueries";
+
+// Retrieved from: https://stackoverflow.com/a/16245768
+const b64toBlob = (
+  b64Data: string,
+  contentType = "",
+  sliceSize = 512,
+): Blob => {
+  const byteCharacters = atob(b64Data);
+  const byteArrays = [];
+
+  for (let offset = 0; offset < byteCharacters.length; offset += sliceSize) {
+    const slice = byteCharacters.slice(offset, offset + sliceSize);
+
+    const byteNumbers = new Array(slice.length);
+    for (let i = 0; i < slice.length; i += 1) {
+      byteNumbers[i] = slice.charCodeAt(i);
+    }
+
+    const byteArray = new Uint8Array(byteNumbers);
+    byteArrays.push(byteArray);
+  }
+
+  const blob = new Blob(byteArrays, { type: contentType });
+  return blob;
+};
+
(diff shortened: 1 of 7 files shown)


frontend/src/utils/FileUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+    case "jpeg":
+      return "image/jpeg";
+    case "doc":
+      return "application/msword";
+    case "docx":
+      return "application/vnd.openxmlformats-officedocument.wordprocessingml.document";
+    case "pdf":
+      return "application/pdf";
+    default:
+      return "application/octet-stream";
```
gaoxk, line 47: is this a good catch-all/default file type? what extension does this end up as?

Puepis: oh so the extension is defined by the `file.ext` field but `application/octet-stream` serves as a default type for unknown file types:

| application/octet-stream is the default value for all other cases. An unknown file type should use this type. Browsers pay a particular care when manipulating these files, to protect users from software vulnerabilities and possible dangerous behavior.

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

it can also be used for `.bin` files

gaoxk: looks good to me, ty for including some further reading!

gaoxk (approved):
NICEEEE

PR #276, opened 2021-12-29
Export story translation .docx

lynn-kim:
## Notion ticket link

[Export Story Translation](https://www.notion.so/uwblueprintexecs/Export-Story-Translation-89e4d88148824aecb5e6e72f843775dc)

## Implementation description

- creates a new word doc, saves to db, returns the encoded string, and then deletes the file 

## Steps to test

1. Login as an admin (angela merkel) 
2. [Try this query](http://localhost:5000/graphql?variables=null&operationName=undefined&query=query%20%7B%0A%20%20exportStoryTranslation(id%3A2)%7B%0A%20%20%20%20file%0A%20%20%7D%0A%7D) - see if that returns an encoded string and that file got deleted
3. To see if the file actually saved correctly, (testing process is a little sus rn but) comment out line 53 in `story_query.py`
4. Get the file id for the document (should be the last row in the files table)
5. Set the resume for Carl Sagan to the new file (`update users set resume=<file_id> where id = 1;`)
6. Go to http://localhost:3000/#/user/1 and download the resume 

- How do we want to format the document?
- What other information do we want to retrieve/add to the doc?
- Not sure if there's a way to use `create_file`

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/story_service.py
+++ b/backend/python/app/services/implementations/story_service.py
@@ -93,25 +93,30 @@ class StoryService(IStoryService):
             raise error
 
     def export_story_translation(self, id):
-        story_details = self.get_story_translation(id=id)
-        doc = docx.Document()
-        doc.add_heading(story_details["title"])
+        try:
+            story_details = self.get_story_translation(id=id)
+            doc = docx.Document()
+            doc.add_heading(story_details["title"])
 
-        for line in story_details["translation_contents"]:
-            doc.add_paragraph(line["translation_content"])
+            for line in story_details["translation_contents"]:
+                doc.add_paragraph(line["translation_content"])
 
-        upload_folder_path = os.getenv("UPLOAD_PATH")
+            upload_folder_path = os.getenv("UPLOAD_PATH")
 
-        upload_path = os.path.join(
-            upload_folder_path, secure_filename("downloadable-story-translation.docx")
-        )
-        doc.save(upload_path)
-        file_dto = FileDTO(path=upload_path)
-        self.logger.info(file_dto)
-        new_file = File(**file_dto.__dict__)
-        db.session.add(new_file)
-        db.session.commit()
-        return new_file.to_dict()
+            upload_path = os.path.join(
+                upload_folder_path,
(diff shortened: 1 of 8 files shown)


backend/python/app/services/implementations/story_service.py
```diff
@@ -87,6 +92,27 @@ def import_story(self, details, file):
             self.logger.error(str(error))
             raise error
 
+    def export_story_translation(self, id):
```
gaoxk, line 95: lets wrap the body of this in a try-except statement to handle any errors that might pop up

backend/python/app/services/implementations/story_service.py
```diff
@@ -87,6 +92,27 @@ def import_story(self, details, file):
...
+
+        upload_folder_path = os.getenv("UPLOAD_PATH")
+
+        upload_path = os.path.join(
+            upload_folder_path, secure_filename("downloadable-story-translation.docx")
+        )
+        doc.save(upload_path)
+        file_dto = FileDTO(path=upload_path)
+        self.logger.info(file_dto)
+        new_file = File(**file_dto.__dict__)
```
gaoxk, line 111: hmmmm is it necessary to make a file table entry if it's going to be deleted? is this here for now for ease of testing?

gaoxk (changes requested):
very nicceeeeee 😎 everything worked as expected. clever way to test it! 
I'll run the exported .docx by the client and see if they want any extra info (translator name etc). We can make another ticket for those changes once the info is confirmed.

backend/python/app/services/implementations/story_service.py
lynn-kim: added it so i could use the get_file function since it uses db query to get the file

gaoxk: hmmm i see, mind extracting out the relevant parts of `get_file` to a different function so you can call that directly? That way we can eliminate unnecessary DB commits

gaoxk (approved):
niceeeeeee. tested resume download, so file_by_id seems to be bueno

PR #320, opened 2022-01-07
add is_rtl query for languages

andrewyguo:
## Notion ticket link

[Ticket ](https://www.notion.so/uwblueprintexecs/a658c933d18c48d4b43fd99bda2bd3b4?v=ba22ac93d09040ff8219c60ac7caabdb&p=91c5a6a68efb4c48a3147c8173d38161)

## Implementation description

- Add new `is_rtl` query for languages 

## Steps to test

1. go to GQL and run [this](http://localhost:5000/graphql?query=query%20IsRtl%20%7B%0A%20%20isRtl(language%3A%20%22ARABIC%22)%20%7B%0A%20%20%20%20isRtl%0A%20%20%7D%0A%7D&operationName=IsRtl), ensure it returns `true`
2. Also test with another language like MANDARIN which should return `false`

- unifying quantum mechanics and general relativity 

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/language_service.py
+++ b/backend/python/app/services/implementations/language_service.py
@@ -26,3 +26,17 @@ class LanguageService(ILanguageService):
         except Exception as error:
             self.logger.error(str(error))
             raise error
+
+    def get_is_rtl(self, language):
+        try:
+            to_check = Language.query.filter_by(language=language).first()
+
+            if not to_check:
+                raise Exception(
+                    "Language {language} not found".format(language=language)
+                )
+            else:
+                return {"is_rtl": to_check.is_rtl}
+        except Exception as error:
+            self.logger.error(str(error))
+            raise error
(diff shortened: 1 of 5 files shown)


backend/python/app/services/implementations/language_service.py
```diff
@@ -26,3 +26,12 @@ def add_language(self, language, is_rtl):
         except Exception as error:
             self.logger.error(str(error))
             raise error
+
+    def get_is_rtl(self, language):
+        try:
+            to_check = Language.query.filter_by(language=language).first()
+
+            return {"is_rtl": to_check.is_rtl}
+        except Exception as error:
```
gaoxk, line 35: i know you have this except but can you add an extra err catch in the try if the language does not exist in the db? eg
image
i just feel like querying the wrong language will happen often enough that a special message for it would be helpful

gaoxk (approved):
works 😎

PR #342, opened 2022-01-17
Sort languages alphabetically in language modal

ArmanyaDalmia:
## Notion ticket link

[Sort languages alphabetically in language modal](https://www.notion.so/uwblueprintexecs/Sort-languages-alphabetically-in-language-modal-c7773d85701d49d8a4b195c2d3101726)

## Implementation description

- Just wrote basic sort function for the language options (Sorts only by alphabet, not case)

## Steps to test

1. Login as admin and add a language on Manage Story Translations page

- Sort function implementation (should I use .toLowerCase() since that is used elsewhere or does it not matter?)

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/language_service.py
+++ b/backend/python/app/services/implementations/language_service.py
@@ -11,7 +11,7 @@ class LanguageService(ILanguageService):
 
     def get_languages(self):
         try:
-            languages = Language.query.all()
+            languages = Language.query.order_by(Language.language.asc()).all()
             return [language.language for language in languages]
         except Exception as error:
             self.logger.error(error)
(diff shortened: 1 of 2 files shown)


lynn-kim (changes requested):
Hm the languages only appear sorted after the admin adds a language, so in the first load or after a refresh the languages appear out of order - one option is to move the sorting to a separate util file and reuse it around line 88 when the languages are fetched

gaoxk (changes requested):
hmmmm I know im flip flopping on my decision but can you instead sort on the backend at the get_lanaguages query? while the languages are sorted here, there not in the filters and elsewhere

lynn-kim (approved):
tested, lgtm!

PR #380, opened 2022-02-25
Debug MySQL Lost Connection Issues

jennchenn:
## Notion ticket link

[MySQL Lost Connection](https://www.notion.so/uwblueprintexecs/MySQL-Lost-connection-c1e7ee9ff2744d68b4c3dc8735d0f9d1)

## Implementation description

- add wrapper around all database calls to:
1. pre-ping database to ensure connection is open
2. catch any errors after commit and roll them back 

## Steps to test

1. does the production database connection error out anymore?

- `backend/python/app/services/implementations/utils.py`
- is there a better way of ensuring connection is valid other than running `SELECT 1;` before each db query? it shouldn't be very expensive but not sure if it'd be preferable to retry once on error instead?

## Checklist

### DIFF
--- a/backend/python/app/__init__.py
+++ b/backend/python/app/__init__.py
@@ -66,6 +66,7 @@ def create_app(config_name):
     )
 
     app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+    app.config["SQLALCHEMY_ENGINE_OPTIONS"] = {"pool_recycle": 1800}
     mysql.init_app(app)
     firebase_admin.initialize_app()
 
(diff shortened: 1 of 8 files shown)


gaoxk (approved):
select 1 is o(1) computationally so im happy with this solution! thanks for powering through it, your a super hero!

backend/python/app/__init__.py
```diff
@@ -66,6 +66,8 @@ def create_app(config_name):
     )
 
     app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+    app.config["SQLALCHEMY_POOL_RECYCLE"] = 499
+    app.config["SQLALCHEMY_POOL_TIMEOUT"] = 20
```
gaoxk, line 70: out of curiosity, did you find that you needed both the select 1 solution and these ping configs?

